PR #15, opened 2017-03-28
Update lesson2.md

IsmaelMartinez:
I  think is important to start talking about "this" as understanding that part of javascript helps understanding the callbacks and the language in general.

Here a small resume of what "this" is:

At the time of execution of every function, JavaScript engine sets a property to the function called this which refer to the current execution context. this is always refer to an object and depends on how function is called. There are 7 different cases where the value of this varies.
* In the global context or inside a function this refers to the window object.
* Inside IIFE (immediate invoking function) if you use "use strict", value of this is undefined. To pass access window inside IIFE with "use strict", you have to pass this.
* While executing a function in the context of an object, the object becomes the value of this
Inside a setTimeout function, the value of this is the window object.
* If you use a constructor (by using new keyword) to create an object, the value of this will refer to the newly created object.
* You can set the value of this to any arbitrary object by passing the object as the first parameter of bind, call or apply
* For DOM event handler, value of this would be the element that fired the event

### DIFF
--- a/js-core/lesson2.md
+++ b/js-core/lesson2.md
@@ -5,6 +5,7 @@
   - Functional utils (map, foreach, filter)
 - Objects
 - Callbacks
+- this
 
 ---
 
@@ -31,4 +32,5 @@ High level explanation of Topic 1 - rely on the exercises
 4. **Research:**
 
 ## Prepare for the next class
-1. Read this [Some Tutorial or Video etc...](https://google.com)
+1. What is "this" [Understanding Javascript this in a crystal clear way](https://youtu.be/yuo8YqKDQ-M)
+2. Read this [Some Tutorial or Video etc...](https://google.com)


rarmatei:
"this" is one of the most hard to understand concepts in JavaScript. If you have a look at the Lesson 4 issue I opened that's where I suggested it should go, as I thought students might need to get a bit more used to JS before diving into more advanced topics.

Furthermore, in this lesson we've just introduced "functions" and "objects", and maybe students might need some time to focus on their homework around these  topics before we make the leap to "this". What do you think?

IsmaelMartinez:
hi @rarmatei , agree that is hard to understand but, if we are talking about callbacks, they should at least get a wee taste of "this", "that" and "self"... even if its just a mention. 

I tend to find people understanding better concepts when they are repeated multiple times over time. 

But agree that is a difficult concept even to explain!!

rarmatei:
@ISMAELMARTINEZ Sure, I fully agree that they need to have a basic idea of "this" by the end of the whole course, I'm just not sure it should be presented this early on?

If you look at Kyle Simpson's [course](https://frontendmasters.com/courses/advanced-javascript/), he doesn't cover the "this" keyword until he gets to the middle of his advanced JavaScript course. And he has loads of year of experience in teaching and writing about JS.

IsmaelMartinez:
Not a problem.

PR #21, opened 2017-04-02
Lesson 0 - Terminal and homework

claudiamatosa:
- Moved "Terminal" to the top, as by the time they reach the Github part they will be reading about commands.
- Added simple homework tasks. I noticed that moving their website from Codepen to Github is part of lesson 1. Should we add it here as a stretch goal?
- Should we recommend only one IDE, so they will all be using the same?


lesson0.md
```diff
@@ -1,15 +1,64 @@
...
+your computer. [More instructions >](http://www.wikihow.com/Open-a-Terminal-Window-in-Ubuntu)
+
+#### Mac OSx (Apple)
+
+The Terminal is already installed. It should be in the list of applications available in
+your computer. [More instructions >](http://www.wikihow.com/Get-to-the-Command-Line-on-a-Mac)
+
+#### Windows
+
+Install [cmder](http://cmder.net/). After that, open it from your windows menu. When
```
claudiamatosa, line 55: Would this be the best tool?

kabaros: not sure - I was thinking Git Bash would do on Windows. I think cmder has a bash shell?

claudiamatosa: Git bash could actually be alright, I had forgotten about it! Will spawn a virtual machine to check how they perform.

PR #22, opened 2017-04-03
WIP: JS lesson 1

tvooo:
This PR is work in progress, I'm going to add to it over time.

Feel free to discuss the contents!


js-core/lesson1.md
```diff
@@ -1,19 +1,176 @@
...
+> If you need single or double quotes *inside the string*, you need to “escape” them by putting a backslash in front of them. If you don’t do this, the JavaScript engine will think the string ends here, because it encounters a quote.
+
+The most common operation on strings is to append one string to another. This is called *string concatenation*. It’s achieved by the plus (`+`) operator:
+
+```
+"Hello" + "World" // "HelloWorld"
+```
+
+- **TODO:** Introduce string concatenation
+- **TODO:** What about basic string functions, like substr, replace etc?
```
rarmatei, line 82: Maybe introduce one of them and give the rest as homework?

js-core/lesson1.md
```diff
@@ -1,19 +1,176 @@
...
+```
+
+Operators have the same precedence as in algebra: `*` and `/` have higher precedence than `+` and `-`.
+
+- **TODO:** Do math exercise
+- **TODO:** What about Infinity, Math.PI, IEEE floating point bullshit, NaN
+
+### Strings
+
+Strings represent any sort of text. They are delimited by single quotes (`'`) or double quotes (`”`) and can be of any length.
```
rarmatei, line 62: Don't know if worth introducing at this stage, but recently I've exclusively started backticks `` for surrounding multi-line strings or strings that contain a lot of variables: `Hello ${userName}!`

js-core/lesson1.md
tvooo: Good idea

js-core/lesson1.md
tvooo: I wasn't sure either. @kabaros what is our approach to ES6+ things like template strings? This also informs if we introduce `const` and `let` early on, which really goes down a rabbit whole of "scope" wayyy to early, I think.

kabaros: @tvooo @rarmatei  I would say leave it out from the first lessons. We will need to introduce those somehow after 3/4 weeks (maybe as a research/homework to start with)

tvooo:
Rebased onto current master and added more things. Please, have a look and let me know if you think my level of detail and the way I introduce things is good / makes sense. Would be a shame to put more work into it if it goes into the wrong direction.
Also, any ideas for non-research homework and for exercises are highly welcome!

PR #31, opened 2017-04-17
moving the homework for HTML lessons to a different repo

kabaros:
I moved the Homework (the part about building a site) to a seprate repo https://github.com/Code-Your-Future/html-css-project

This is mainly to get them to fork it, start clean (to avoid having these forks out of lessons as in the last class) and build on the fork over the three weeks.

One more minor thing is that I added a link to a w3c explaining how HTML is strucutured - we should spend less than half hour on that in the class as a quick review but it would be worth doing for people who are not very comfortable with HTML.

@NateWr .. If you look at the Home work repo, I've also added an image of a Blog sketch for them to follow it while developing their sites. Please have a look at that as well.

#5


claudiamatosa (approved):
Looks good. 👍

Minor things:
- I'd add a link with help on how to fork a repo, as some of the students are not very familiar with Github yet and `fork`. Although some of them will discover it on their own, others will need more guidance.
- In the homework itself, would add a note saying that this is just a sketch - they can choose the colours, fonts and images that they want to use in the page.

kabaros:
@claudiamatosa added a note about forking - I will update the homework as well.

PR #33, opened 2017-04-22
HTML Lesson 1: Use Github instead of CodePen for bikes for refugees

claudiamatosa:


html-css/lesson1.md
```diff
@@ -25,7 +25,9 @@ In today's class, we will begin adapting styles on this example website. We'll r
   <img src="assets/screenshot-complete.png" style="border: 1px solid #bababa; width: 48%;">
 </a>
 
-The example website you'll begin working with is available [on this Codepen](https://codepen.io/natewr/project/editor/AwpkjD/).
+The example website you'll begin working with is available [on Github](https://github.com/Code-Your-Future/bikes-for-refugees).
+
+In order to get started, please [fork it](https://help.github.com/articles/fork-a-repo/) on Github, so it shows up under your account. Then clone it in your terminal and open it with VS Code (our chose IDE).
```
nennes: Maybe clone could be a link like this? `https://help.github.com/articles/cloning-a-repository/`
Also, `our chose IDE` is probably wrong, maybe something like `our chosen IDE` is what you meant?

claudiamatosa: Ahaha, committing too fast!

PR #36, opened 2017-04-23
HTML Lesson 1: Optional homework

claudiamatosa:


docs/react/week-1/lesson.md
```diff
@@ -110,13 +91,15 @@ ReactDOM.render(element, rootElement);
...
+| :----------------------------------------------------------------------------------------------------------------------------------------------------- |
+| Change the [JSX example from above](http://jsbin.com/gekahexige/edit?html,output) to instead render a `h1` tag with the text "Hello Code Your Future". |
 
 ## Let's create a React app
 
-> **Exercise C:**
-> If you haven't already, follow [the instructions to create a React app] called `pokedex`(https://docs.codeyourfuture.io/students/guides/creating-a-react-app).
+| Exercise C (estimate: 10 min)                                                                                                                                     |
+| :---------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| 1. If you haven't already, follow [the instructions to create a React app] called `pokedex`(https://docs.codeyourfuture.io/students/guides/creating-a-react-app). |
```
illicitonion: This link isn't formatted quite right I think

docs/react/week-1/lesson.md
```diff
@@ -145,16 +128,16 @@ There are 3 important parts in this code:
 2. We create a React component called `HelloWorld`.
 3. We _render_ the `HelloWorld` component into a `div` with the id of `root`.
```
illicitonion:
```suggestion
3. We _render_ the `HelloWorld` component into the element with the id of `root`.
```

There _is_ a `div` tag in this code, but there's nothing in the code which shows that `root` is a `div`, so I think it's worth removing the `div` in this sentence to avoid possible confusion.

docs/react/week-1/lesson.md
```diff
@@ -145,16 +128,16 @@ There are 3 important parts in this code:
...
+| 2. Delete everything in the file except the line containing `export default App`. You should see an error in your terminal and in your web browser - don't panic! We're going to remake the `App` component ourselves.                    |
+| 3. Import the React variable from the React package.                                                                                                                                                                                      |
+| 4. Create a function named `App`, which will be our component.                                                                                                                                                                            |
+| 5. Within the `App` function, return a `<h1>` element with the text "Welcome to the Pokedex". What do you see in your web browser?                                                                                                        |
+| 6. Create a `<div>` element that _wraps around_ the `<h1>` you just created.                                                                                                                                                              |
```
illicitonion, line 142: Is it worth explaining _why_ we need to do this (i.e. that JSX assumes it has a single root element)? I've seen a lot of students get confused because they didn't have a wrapper element, and the error message wasn't super clear to them

docs/react/week-1/lesson.md
```diff
@@ -185,14 +168,14 @@ We are also using some shorter syntax within the `HelloWorld` component. `<Greet
 
 Notice how the components that we write (`HelloWorld`, `Greeting`, `Mentor`) are written using a `camel case` convention and always start with an uppercase letter? And "regular DOM" components (`div`, `span`) are always lowercase? This the convention to let you know whether you are using a "regular DOM component" or a component that you have written.
```
illicitonion:
```suggestion
Notice how the components that we write (`HelloWorld`, `Greeting`, `Mentor`) are written using a `camel case` convention and always start with an uppercase letter? And "regular DOM" components (`div`, `span`) are always lowercase? This is a convention to let you know whether you are using a "regular DOM component" or a component that you have written. When you're making your own components, you should always start them with an uppercase letter.
```

docs/react/week-1/lesson.md
```diff
@@ -185,14 +168,14 @@ We are also using some shorter syntax within the `HelloWorld` component. `<Greet
...
+| 1. In your `pokedex` React app, open the `src/App.js` file.                                                                                                                        |
+| 2. Create a new function named `Logo`.                                                                                                                                             |
+| 3. Copy the `<header>` element and its contents and paste it into the `Logo` component.                                                                                            |
+| 4. Replace the `<header>` element in the `App` component with the new `Logo` component.                                                                                            |
+| 5. Create a new component function named `BestPokemon` and return a `<p>` element with some text saying which is your favourite Pokemon (e.g. "My favourite Pokemon is Squirtle"). |
+| 6. _Render_ your new `BestPokemon` component below the `Logo` component within the `App` component.                                                                                |
 
 #### Arrow Functions for shorter syntax
```
illicitonion, line 184: In the block below, it says "This can be even shorter again if we use parentheses and implicit return:" - do you know what it means by "use parentheses"? I'm not sure, so it may be worth rewording to clarify... I _think_ it's trying to get a "if your JSX goes on more than one line, it needs to be in parentheses", but the example it uses doesn't do so.

docs/react/week-1/lesson.md
```diff
@@ -423,31 +407,30 @@ Or calculating new values:
...
+| 9. Paste the `abilities` variable into `App.js`.                                                                                                                              |
+| 10. Pass the `abilities` variable as a prop to `BestPokemon` from `App.js`.                                                                                                   |
+| 11. In the `BestPokemon.js` file replace the existing usage of `abilities` with the `abilities` **prop**. You should still see the Pokemon ability names in your web browser. |
+| 12. **(STRETCH GOAL)** Repeat the process with the `date` variable in the `CaughtPokemon.js` file.                                                                            |
 
 ## Further reading
 
 Fed up of having to provide a prop for every component? Do you want to make your component use a value most of the time, but it can be overridden with a prop? This is a good time to use `defaultProps`. [This page on the React documentation](https://reactjs.org/docs/typechecking-with-proptypes.html#default-prop-values) describes in more detail.
```
illicitonion: This page is purely class-focused - I'm not sure the small benefit of `defaultProps` is worth the confusion... I guess maybe for those who are eager enough to do further reading it'll be useful/interesting? Maybe it's worth throwing in a sentence saying something like "This example uses classes, which are different ways to define functions. We're not going to use classes in this course, but it's good to know that if you see some code which looks like:
```js
class MyComponent {
...
}

MyComponent.cats = "cute";
```
when you're using functions, you can write:
```js
function MyComponent {
...
}

MyComponent.cats = "cute".
```

docs/react/week-2/lesson.md
```diff
@@ -95,11 +87,9 @@ function ClickLogger() {
 }
 ```
 
-You might find it a little strange that we have a function inside a function. But this is a normal thing to do in JavaScript! `logWhenClicked` is within the _scope_ of our `ClickLogger` component.
```
illicitonion, line 98: In the block above this: "We still need to listen events in React" should probably be rephrased slightly

docs/react/week-2/lesson.md
```diff
@@ -110,11 +100,13 @@ Here's a few examples:
 - `onPlay` - a video starts playing
 - `onSubmit` - a form element is submitted
 
-A full list of special event handler props is available [here](https://reactjs.org/docs/events.html#reference).
+Notice that just like with `addEventListener` above, we pass the function reference to `onClick` instead of calling the function. If we call the function, it will run when we render, not when the user clicks on the button.
```
illicitonion: I'm not sure we've actually defined/explained the word "render" in week 1. I think it's implicitly meant to be explained in the sentence "We _render_ the HelloWorld component into a div with the id of root." in the "React Components" section, but it may be worth adding a 1-2 sentence explanation there (and maybe re-capping it here).

docs/react/week-2/lesson.md
```diff
@@ -259,11 +251,28 @@ const [count, setCount] = useState(0);
 
 To fully understand this bit of code, we first have to understand _destructuring_. Let's look at [this blog post by Wes Bos about array destructuring](https://github.com/wesbos/es6-articles/blob/master/19%20-%20Destructing%20Arrays.md).
 
-Now we can understand that `useState` is returning an array, with two items. The first item in the array is the current value of the `count` state. In our example it will be 0 on the first render.
+Now we can understand that `useState` is returning an array, with two items. The first item in the array is the current value of the `count` state. In our example it will be 0 on the first render. The second item in the array is a function that we will use to update our state.
```
illicitonion, line 254: In the block above, "Finally, let's at how we get hold of the count variable:" is missing probably the word "look"

docs/react/week-2/lesson.md
```diff
@@ -316,12 +325,14 @@ function Counter() {
...
-| :----------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| 1. Open the `pokedex` React application and open the `CaughtPokemon.js` file.                                                                                |
-| 2. Create a new state variable called `totalCaught` and initialise it to `0`                                                                                 |
-| 3. When you create the `totalCaught` state, you should also set the function that will update this state (hint: refer to the syntax of the `useState` hook). |
-| 4. Replace the number 0 in the JSX with your new `totalCaught` state.                                                                                        |
+| Exercise D (estimate: 15 min)                                                                                                                                                                                                                                                                                                                                                                 |
```
illicitonion, line 328: I'd possibly be a bit more explicit in the "When re-rendering" sentence above; something like:

When re-rendering, React will call our `Counter` component again, but this time when we call `useState` it will give us the **updated** state of `1`, instead of the _initial_ state of `0`.

docs/react/week-2/lesson.md
```diff
@@ -316,12 +325,14 @@ function Counter() {
...
+| 3. Within the JSX, there should be a "hard-coded" number 0. Replace it with your new `caught` state.                                                                                                                                                                                                                                                                                          |
+| 4. Add a button to the component with an `onClick` handler that calls a function called `catchPokemon`.                                                                                                                                                                                                                                                                                       |
+| 5. Create the `catchPokemon` function and update the `caught` state so that it is increased by 1 on each click. <details><summary>Click here if you are stuck.</summary>You will need to call the set state function (the 2nd item in the `useState` array) with `caught + 1`.</details>                                                                                                      |
```
illicitonion:
```suggestion
| 5. Create the `catchPokemon` function and have it update the `caught` state so that it is increased by 1 on each click. <details><summary>Click here if you are stuck.</summary>You will need to call the set state function (the 2nd item in the `useState` array) with `caught + 1`.</details>                                                                                                      |
```

docs/react/week-2/lesson.md
```diff
@@ -316,12 +325,14 @@ function Counter() {
...
+| 6. Write down the things that will happen when you click the button. Compare your list with another student and discuss. <details><summary>Click here for a hint.</summary>The state will be updated to be the current state + 1. React is notified that our state has changed, so it re-renders. When rendering, the current state will be different and so React updates the DOM.</details> |
```
illicitonion, line 335: The wording of the "Don't mutate state" section implies that React is throwing an error, or at least logging a warning somewhere. The nuance of making something readonly is a bit more fiddly than I think we should go into, though... I'd maybe rewrite it as:

As we just learned, `setCount` updates the state for us, but it also notifies React of changes. If you try to just change the `count` variable without using `setState`, nothing will happen, because React wouldn't be notified of the change. You can _only_ modify state using the setter function.

And then maybe removing the sentence "React is letting us know that we have to use `setCount` to be able to update state."

docs/react/week-2/lesson.md
```diff
@@ -362,49 +373,90 @@ function App() {
...
+      <button onClick={addToList}>Add item to list</button>
+      {list.map((item, index) => {
+        return <li key={index}>{item}</li>;
+      })}
+    </p>
+  );
+}
+```
+
+In this shopping list example, we're initialising the `list` state to be an empty array. To display our list we loop through the array (like we learned last week) and render each item on the list.
```
illicitonion:
```suggestion
In this shopping list example, we're initialising the `list` state to be an empty array. To display our list we loop through the array (like we learned last week) and render an `<li>` for each item in the list.
```

docs/react/week-3/lesson.md
```diff
@@ -134,7 +272,7 @@ Together let's "play computer" to break down exactly what is happening with thes
   }
 ```
 
-8. Now `App` **does** render `MartianImageFetcher` and passes the `date` state as a prop (also named `date`)
+8. Now `App` **does** render `MartianImageFetcher` and passes the `date` state as a prop named `photoDate`
 9. The `MartianImageFetcher` function component is called for the first time
 10. `useState` knows that we haven't got any `imgSrc` state so initialises it to `null`
 11. We queue an effect, which will run after we render for the first time
```
illicitonion, line 287: On the next line, I think it's the first time we're explained that returning `null` renders nothing. Could we explain this earlier (maybe in the ternary/&& section), so that this isn't _another_ piece of new information in this already quite complex set of bullet points?

docs/react/week-3/lesson.md
```diff
@@ -246,13 +384,18 @@ Phew! That was a lot of work just to render an image! But we're not quite done y
...
-| **Exercise B**                                                                                        |
+| Exercise D (estimate: 5 min)                                                                          |
 | :---------------------------------------------------------------------------------------------------- |
 | 1. Did you spot where the bug was? Discuss with a group of 2 - 3 students where you think the bug is. |
 | 2. Report back to the rest of the class where you think the bug happened.                             |
 
+<details>
+<summary>Click here to reveal the answer!</summary>
+
 The key that the `useEffect` in `MartianImageFetcher` is **only run once**. This is because we told React that the queue should be queued on the first render only. However, as we saw, sometimes you need the effect to run again when some data changes. In this case the `date` prop, changed from `"2019-01-01"` to `"2020-01-01"`, meaning that we have to fetch data different data. We call this a _dependency_ of the effect.
```
illicitonion:
```suggestion
The key that the `useEffect` in `MartianImageFetcher` is **only run once**. This is because we told React that the queue should be queued on the first render only. However, as we saw, sometimes you need the effect to run again when some data changes. In this case the `date` prop, changed from `"2019-01-01"` to `"2020-01-01"`, meaning that we have to fetch data different data. We call this a _dependency_ of the effect - any of our variables which we use in the callback function we gave to `useEffect` is a dependency.
```

PR #44, opened 2017-05-05
Update lesson 2 exercises

nennes:
Thoughts on the lesson 2 exercises, needs some more work.


claudiamatosa:
Exercises looks good! Suggestions:

- `Content layout: floats (maybe), Content layout: relative and absolute positioning (maybe)` - if we are going to leave these items here, I’d remove the maybe. Should we remove the `floats` part altogether, as it’s a practice falling quickly out of usage, and leave `relative and absolute positioning`? We’ll need to add a couple of exercises for that as well.
- “jumbotron” is quite a big work and I’m not sure if students will understand it. They don’t have our Googling experience and may just feel lost when they see a word they don’t understand. I’d suggest adding an explanation between parenthesis, or changing the wording altogether. Or adding an example with images!
- Would add examples with images, so they know what we’re looking for.
- `Exercise: Use Flexbox for the layout of the rest of your bikes for refugees website.` - the `upcoming events` section is broken in the html we gave them initially. Instead of being vague, we can ask them to make it look like the original picture using flexbox (we definitely need to add image examples here!).

<table>
<tr>
<td valign="top">
<h3>Current:</h3>
screen shot 2017-05-06 at 13 12 28
</td>

<td valign="top">
<h3>Result we're looking for:</h3>
screen shot 2017-05-06 at 13 13 17
</td>
</tr>
</table>


- For the next class, we’ll want the guys to start building their own library. I’d  remove point 4 from the homework, so it isn't too much, and update the prepare for next class with instructions about the library. It needs to use simple language and lots of examples. What do you think? I can do that while you update the content.

PR #55, opened 2017-05-14
HTML-CSS lesson 3 - fix typos

claudiamatosa:


docs/js-core-3/week-3/lesson.mdx
```diff
@@ -287,319 +285,168 @@ function six() {
...
-
-const runningTotal = new RunningTotal();
-
-runningTotal.add(3);
-runningTotal.add(7);
-runningTotal.add(7);
-runningTotal.getTotal();
-// 17
 ```
+let student = ['Sara','Lee', 30, 'London', 'Javascript'];
```
alismith-dev: This exercise feels a bit contrived as an object would be the better data structure here.

How about switching to something like this:

```js
let familyMembers = ['Homer', 'Marge', 'Bart', 'Lisa']

console.log(`In the Simpsons family, ${dad} and ${mum} are the parents. ${son} is the son, and ${daughter} is the daughter`)
```

?

This also would emphasise the ordered nature of destructuring.

docs/js-core-3/week-3/lesson.mdx
```diff
@@ -287,319 +285,168 @@ function six() {
...
-  console.log(`Button has been clicked ${timesClicked2} times`);
-};
-
-document
-  .querySelector("#myButton1")
-  .addEventListener("click", whenButtonClicked1);
-document
-  .querySelector("#myButton2")
-  .addEventListener("click", whenButtonClicked2);
+let fruits = ['apple', 'orange', 'banana'];
```
alismith-dev: Could we mark these snippets as `js`? We would get syntax highlighting then, and I think would format them with Prettier (e.g. I believe it would switch to double-quotes).

docs/js-core-3/week-3/lesson.mdx
```diff
@@ -287,319 +285,168 @@ function six() {
...
-class Person {
-  constructor() {
-    this.name = "Mike";
-  }
-
-  sayHi() {
-    console.log(`Hi, ${this.name}`);
-  }
-}
+Which of the following lines of code can be put on line 5?
```
alismith-dev: Should we make this (and perhaps the A/B/C/D below) a comment?

docs/js-core-3/week-3/lesson.mdx
```diff
@@ -287,319 +285,168 @@ function six() {
...
-  sayHi = () => {
-    console.log(`Hi, ${this.name}`);
-  };
-}
-
-const person = new Person();
-const sayHi = person.sayHi;
-
-sayHi(); // Hi, Mike
+D) let [sum, multiply] = sumAndMultiply(2, 3);
```
alismith-dev:
```suggestion
// Option D
let [sum, multiply] = sumAndMultiply(2, 3);
```

docs/js-core-3/week-3/lesson.mdx
```diff
@@ -287,319 +285,168 @@ function six() {
...
+#### Exercise (4)
 
-```js
-// Exercise 1
-class Car {
-  goFaster() {
-    this.speed = 100; // what is the value of this code below is run
-  }
-}
+Complete all of the exercises found [here](https://repl.it/@ChrisOwen101/ArrayDestructing)
```
alismith-dev: In my experience, it is helpful to move exercises on external services into a "team" (if possible), so that if changes are desired, we don't need to update the link.

Looks like there is a free tier for the repl.it teams: https://repl.it/teams. Might be worth a quick email to see if we can get a paid account?

alismith-dev: To comment on the exercises themselves, I think they suffer a bit from being a bit contrived, similar to the exercise above.

Several of them feel like they would be better suited to objects than arrays, which I think will either be confusing, or lead to bad practice. In general I would suggest trying to find alternatives where there is some kind of order to the data.

Having said that, `useState` somewhat breaks this idea as there isn't an inherent reason that it gives the current state followed by the state setter. Array destructuring was only used by the React team as it allows us to name the values whatever we like. So perhaps some examples of this wouldn't be too bad.

Sorry that I don't have more concrete suggestions.

alismith-dev: Oh and I think you've misspelled destructuring in the repl.it title 🙂

ChrisOwen101: Looks like we already have a CodeYourFuture team since that name is taken - I'll move it to that account as soon as I have access 🙄

docs/js-core-3/week-3/lesson.mdx
```diff
@@ -287,319 +285,168 @@ function six() {
...
-    console.log(`Hi, ${this.name}`);
-  }
-}
+let [firstWord, secondWord, thirdWord, fourthWord] = sentence;
 
-const person = new Person();
-person.sayHi();
+console.log(
+  `FirstWord: ${firstWord}, SecondWord: ${secondWord}, ThirdWord: ${thirdWord}, FourthWord: ${fourthWord},`
+);
```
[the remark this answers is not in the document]
louisechow: Added this as Exercise 4

docs/js-core-3/week-3/lesson.mdx
louisechow: Commented these lines

docs/js-core-3/week-3/lesson.mdx
louisechow: Done.

docs/js-core-3/week-3/lesson.mdx
louisechow: Done.

PR #56, opened 2017-05-14
js-core lesson 1 updated based on Node Girls workshop #6 

kabaros:
this is basically porting the Node Girls Beginners' workshop to our syllabus with some amends. I am still going to go through it to change few things (thinking of using Node directly to run code, instead of repl.it or developer console or anything)

The Node Girls workshop is a great immersive start to JS: I've amended some bits to use more accurate terminology, I also updated some of the code samples to be more relevant to our students. I changed the Array explanation to use Tim's explanation as it is easier to grasp. I've also moved the existing content of Lesson 1 to Lesson 2 as it will serve as a good way to reiterate the more complex concepts like arrays, objects etc...

### DIFF
--- a/js-core/lesson1.md
+++ b/js-core/lesson1.md
@@ -4,198 +4,425 @@
 
 **What we will learn today?**
 
-- Variables, assignment and comparison
-- Conditionals: If/else, switch, ternary operator
+- Hello World
+- Data  Types
+- Variables
+- Conditionals: If/else
+- Loops
 - Arrays
-- Loops: for, while
 - Objects
+- This class is based on the [Node Girls JS Workshop](https://github.com/node-girls/beginners-javascript)
+
 
 ---
 
-## Conditionals
+# Getting Started
 
-**TODO**
+1. Download and install the latest Current version of NodeJS - from https://nodejs.org/en/download/current/
 
-### Comparisons
+To test that it was installed and running properly, go to your terminal and run the command:
+`node -v`
+You should get the node version printed on your terminal, for example, `v7.3.0`
 
-**TODO**
 
-- Introduce `==` and `!=`, `<` and `>`
-- Introduce the concept of truthiness and falsiness for non-boolean types
-- Introduce `===` and `!==`
+2. Fork the repositroy - https://github.com/Code-Your-Future/JS-Core-1-Exercises and **Clone** it locally.
 
-### If/Else
+3. Once you have the repo locally, go to the folder on your terminal and run `node main.js`
 
-**TODO**
+You should get **Hello World** printed on your Terminal.
 
-### Switch
+# Hello World
 
-**TODO**
+It is programming tradition that the first thing you do in any language is make it say "Hello world!". This is the first thing we'll do in JavaScript, using something called a `console.log()`.
 
-### The ternary operator
+## `console.log()`
 
-**TODO**
+When we ran the command `node main.js` - that did **log** to the **console** printing the words **Hello World** to the terminal. 
 
-## Arrays
+Let's open the file **main.js** and investigate its contents
 
-If you want to store a bunch of things, identifiers like `name1`, `name2`, `name3` and so on quickly become tiresome. **Arrays** to the rescue!
+> **Exercises**
+> - What do you notice about the code? Anything you don't understand? Notice the semicolon, the dot, the parantheses - these are all elements of JavaScript **Syntax**
+>
+> - Try to `console.log()` something different, for example, `Hello World. I just started learning JavaScript!`. (run `node main.js` after each change and look at the output in the console)
+> - Try to `console.log()` several things at once.
+> - What happens when you get rid of the quote marks?
+> - What happens when you `console.log()` just a number without quotes?
 
-Imagine an array like a list of variables, but instead of each variable having its own name, only the array has a name. So how do you access a variable inside, then? How do you read and write? Using an *index*!
+## Comments
 
-An index is just a whole number, such as `0`, `1`, `2` and so forth. As in many areas of computing, indices in JavaScript arrays start with `0` --- so `0` is the index of the first element of an Array, `1` is the index of the second element, and so forth.
+You can put comments to help you and others understand your code - use `//` to add a comment (or if it is a multi-line comment, then use `/*` and `*/` around your comment)
+```js
+// This is a comment for myself and other developers to understand the next line of code
+console.log('Hello world!');
 
-### Creating an array
+console.log('Hello World'); // comments can also be in the same line like this
+/*
+I can also make comments
+over more than one line if necessary
+*/
+console.log('Hello World');
+```
+> **Exercises** What are the comments in that `main.js` script.
 
-The easiest way to create an array is using the bracket (`[]`) syntax. Just put whatever elements you want in your array between the brackets, separated by comma:
+## Node?
+We ran our JavaScript using `Node` -  Node is is an open-source, cross-platform JavaScript run-time environment for executing JavaScript code server-side. [Wikipedia](https://en.wikipedia.org/wiki/Node.js)
+> **Exercises** What are other "places" where we can **run** JavaScript code?
+
+# Data Types
+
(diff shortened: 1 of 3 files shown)


js-core/lesson1.md
```diff
@@ -4,198 +4,427 @@
...
+```
+
+
+You've seen what they look like, now make sure they are what you think they are. There is a built-in feature in JavaScript which allows you to check the type of a particular value: `typeof`.
 
-// Remember, the first index is 0
-console.log(countries[0]); // 'Scotland'
-console.log(countries[1]); // 'Germany'
-console.log(countries[2]); // 'Syria'
+Write this code in your Repl, and click "run":
```
NateWr: "Repl"

js-core/lesson1.md
```diff
@@ -4,198 +4,427 @@
...
 ```
 
-You can also create new array elements by just writing to the respective index.
+It basically translates to "if this condition is true, do this - or else do something else".
+
+### Try it out
+
+What if we want to `console.log()` "Hello world!" if we're feeling happy, and something else if we're sad?
+
+Let's define a variable called `isHappy`, and assign it a boolean value of `true` or `false` (depending on how happy you're feeling).
```
NateWr, line 175: `isHappy` is used here, but `happy` is used below.

js-core/lesson1.md
```diff
@@ -4,198 +4,427 @@
...
+if (happy === true) {
+  // if I am happy, print "Hello world!"
+  console.log('Hello world!');
+} else {
+  // if I am sad, print a frowny face
+  console.log(':(');
+}
+```
+
+Now run this code (not forgetting to put the variable at the top of the Repl), and see what happens! Try changing the value of your variable from `true` to `false`.
```
NateWr: "Repl"

js-core/lesson1.md
```diff
@@ -4,198 +4,427 @@
...
-var index = 2;
-console.log(countries[index]); // 'Syria'
+> Write an if/else statement that evaluates if a number is even or odd. If it is even, it will print the string `'even'`, and if it is odd, it will print `'odd'`.
+
+> For this challenge, you might find it useful to use the "modulo" operator, which looks like this: `%`. The modulo operator finds the remainder after division of one number by another. For example:
+
+> ```js
+13 % 2 === 1 // 13 divided by 2 leaves a remainder of 1
+100 % 10 === 0 // 100 divided by 10 leaves a remainder of 0
 ```
```
NateWr: Formatting

selection_257

js-core/lesson1.md
```diff
@@ -4,198 +4,427 @@
...
 
-1. [Resource 1 to learn more about Topic 1](https://google.com)
-2. [Resource 2 for Topic 2](https://google.com)
+> **Exercise**
+Using an object representing a person, `console.log()` a sentence introducing the person. Print out the following:
 
-# Homework
+> ```js
+"Hi, my name is {firstName} {lastName}. I am {age} years old, and work as a {occupation}."
+```
```
NateWr: Formatting
selection_258

js-core/lesson1.md
```diff
@@ -4,198 +4,428 @@
...
-### If/Else
+3. Once you have the repo locally, go to the folder on your terminal and run `node main.js`
 
-**TODO**
+You should get **Hello World** printed on your Terminal.
 
-### Switch
 
-**TODO**
+Fork the repositroy - https://github.com/Code-Your-Future/JS-Core-1-Exercises and **Clone** it locally.
```
rarmatei: This can be removed, you already mention this above

js-core/lesson1.md
```diff
@@ -4,198 +4,428 @@
...
+> - What happens when you `console.log()` just a number without quotes?
+
+## Comments
+
+You can put comments to help you and others understand your code - use `//` to add a comment (or if it is a multi-line comment, then use `/*` and `*/` around your comment)
+```js
+// This is a comment for myself and other developers to understand the next line of code
+console.log(typeof 'Hello world!');
+
+console.log(typeof 'Hello World'); // comments can also be in the same line like this
```
rarmatei: Hmm, was it intentional to add `typeof` here? Might confuse people, as the point of this topic is just to introduce comments?

js-core/lesson1.md
```diff
@@ -4,198 +4,428 @@
...
+When you code, you'll want to create *shortcuts* so you can easily refer to things without writing out the same value every time. In JavaScript we have **variables**, which allow us to store values to use later.
+
+> Think of variables names like **labels** on boxes, while the value of the variable are the **contents** of the box - you could change the contents of a box and leave the label intact, the contents of the boxes can have different types, the boxes should have good labels (a box of books being labeled pens would be very confusing),
+>
+> ![Variables are like boxes](http://2.bp.blogspot.com/-guGunS9ljms/Ui-sF7Q9yQI/AAAAAAAAACU/DdnRgAARYyI/s1600/box.png)
+> Photo from [Khan Academy](http://cs-blog.khanacademy.org/2013/09/teaching-variables-analogies-and.html)
+
+## Try it out
+
+You create a variable using the keyword `var`, which lets the application know that you're about to store a value. You also have to give a name to your variable, which can be anything you like, so long as it doesn't include spaces.
```
rarmatei, line 123: Crazy idea, but hear me out: shall we switch all references to `var` with `let`?

`let` is exactly like `var` but does scoping right, and also properly throws an error if you try to access a variable before it's been declared.

js-core/lesson1.md
kabaros: nope. Will take it out.

js-core/lesson1.md
kabaros: They won't hit these problems at lesson 1, and the majority of examples online would use var, I think it would be confusing if we just skipped over var immediately.

Could we introduce const and let straight away in lesson 2 and encourage them to use them from that point on? (and if so, how do we ease the confusion about the existence of two "versions" of JS)

js-core/lesson1.md
```diff
@@ -4,198 +4,428 @@
...
+>
+> Write an if/else statement that evaluates if a number is even or odd. If it is even, it will print the string `'even'`, and if it is odd, it will print `'odd'`.
+>
+> For this challenge, you might find it useful to use the "modulo" operator, which looks like this: `%`. The modulo operator finds the remainder after division of one number by another. For example:
+>
+> ```js
+>13 % 2 === 1 // 13 divided by 2 leaves a remainder of 1
+>100 % 10 === 0 // 100 divided by 10 leaves a remainder of 0
+>```
+
```
rarmatei, line 215: I'll try to think of another exercise we can add here and submit a PR.  I suspect this whole lesson will be hard for people who've never done programming before, but very easy and hopefully not too boring for people who've programmed before. So I'll see if I can add some more advanced exercises here and there for students who finish early.

js-core/lesson1.md
```diff
@@ -4,198 +4,428 @@
...
+console.log(animals[1]);
+```
+
+Take a guess at which animal that will print out, and see if you're right.
+
+Spoiler alert, it will print out "puppy". That might seem weird, until you learn that JavaScript is a "0-index" language. What that means is that counting in JavaScript starts at 0. So `animals[0]`will print out "tiger". We don't make the rules...
+
+> **Exercise:** **Don't ask a mentor - Google it!**
+How would you `console.log()` the name of every animal in the array?
+If you're thinking a for loop, you're right. This is the perfect opportunity to use our newfound for loop skills for a higher purpose.
```
rarmatei, line 357: `If you're thinking a for loop, you're right.` -> should we avoid telling them what to use, and give them more time to think?

rarmatei:
Looks great!

Apart from my minor comments, I'll have a look this week at adding some more advanced exercises for the more experienced students.

js-core/lesson1.md
rarmatei: I was not hinting at explaining let and const, I think that's a more advanced topic, even for lesson 2. I was just wondering if we can just use let from the beginning, as it's a slightly better version of `var`, and more consistent with scoping behaviour in other programming languages, and couldn't really think of any drawbacks to using it.

But that's a very good point about the online resources. Let's just use `var`, you're right.

js-core/lesson1.md
kabaros: shall I just say you "should loop" .. not mentioning **for**?

js-core/lesson1.md
kabaros: changed it all to isHappy

Issue #10, opened 2017-03-25
JS Core - Lesson 2


kabaros:
https://code-your-future.github.io/syllabus/js-core/lesson2.html

Please mention this issue# in any PRs, commits related to Lesson 2 in JS Core.

rarmatei:
@tvooo We'll be covering Objects and Arrays/Loops in both lesson 1 (which you're building) and 2 (I'm building), so I've been thinking how to best split the topics into two lessons. Here's what I had in mind:

### Objects

- First lesson: introduce objects, explain properties, keys, how to assign/access values
- Second lesson: the `new` keyword, creating objects with constructor functions, primitive types vs reference types

### Arrays, Loops

- First lesson: introduce arrays, how they're different from objects, how their indexes are the keys, and talk about the different types of loops, difference between `do...while` and `while`, `for` loop etc.
- Second lesson: do different exercises that can be solved using loops, as students need to wrap their heads around them and how they can be used to solve more complex problems (find biggest number in an array, then find biggest number out of a series of objects that have a number array property etc.) Maybe as the hard homework they can write an implementation of bubble sort.

What do you think? Would this be a good split of introduce in first lesson, dig deeper in second? Is it too much material?

PR #63, opened 2017-05-24
updated lesson 2 

kabaros:
This is a plan for lesson 2 - it's mainly what's left over from Lesson 1 (arrays and objects). We need to spend as much time as possible honing these basic concepts - in addition, to concepts from the previous week (for loops and functions especially) during this class - some people are still confused about all these abstractions and about what we're trying to do in general.

I suggest we cover the class materials as soon as possible then spend the rest of the time doing the exercises that were supposed to be the homework (https://github.com/Code-Your-Future/JS-Core-1-Exercises) in the class - We should all add more to that repo btw to cover different levels.

The PR includes a basic mentor's guide - it's based on our experience in the first class in Scotland, please go through it, and let me know if anything doesn't make sense.

ps: we probably need a plan to keep the advanced students from getting bored.

#10

### DIFF
--- a/html-css/coding-101.md
+++ b/html-css/coding-101.md
@@ -6,6 +6,7 @@
 
 - What is a computer program?
 - Getting started with JavaScript
+
 ## What is a computer program?
 
 > Intro to programming logic – Making a sandwich (Source: http://static.zerorobotics.mit.edu/docs/team-activities/ProgrammingPeanutButterAndJelly.pdf)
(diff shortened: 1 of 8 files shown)


ash-leigh:
I think this all makes senses and will look into more advance array and object challenges to give the students that did manage the homework - or did none of them manage it?

ash-leigh (approved):

kabaros:
@ash-leigh I think Mohamed Rateb and Khalid did - maybe Yohannes and the other khaled as well

PR #73, opened 2017-06-11
Add exericses for closures and context. Add further resources

dmitrigrabov:


40thieves:
@nbogie raised ([on Slack](https://codeyourfuture.slack.com/archives/C76GQDRH9/p1601380619004600?thread_ts=1601283195.001900&cid=C76GQDRH9)) that setting levels of mandatory homework is challenging, as individual students will be at different levels, and thus should be set different levels of homework (please correct me if I've summarised that incorrectly Neill).

I would propose that we should aim to recommend a certain level of homework, and teachers can adapt as necessary.

PR #87, opened 2017-06-28
Object Oriented Programming intro

kabaros:
skeleton and topics to cover

### DIFF
--- a/js-core-2/lesson5.md
+++ b/js-core-2/lesson5.md
@@ -3,8 +3,7 @@
 # JavaScript Core 5
 **What will we learn today?**
 - Functions and  Good Design
-- Callbacks
-- Objects
+- Object Oriented Programming
 ---
 
 # Functions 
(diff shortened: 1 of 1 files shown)


js-core-2/lesson5.md
```diff
@@ -33,21 +32,107 @@ Find all the bad design issues with this function
 > Exercise
 We have some problem to solve, write really bad code that gives the right result. Tell us why it is bad.
 
+# Object Oriented Programming
+
+**The contents of this section are based on the MDN articles mentioned inthe resources**
+
+> What are abstractions?
+
+In software engineering and computer science, abstraction is a technique for arranging complexity of computer systems. It works by establishing a level of complexity on which a person interacts with the system, suppressing the more complex details below the current level.
```
djgrant, line 41: This is a great definition but I had to read it a few times to get it.

How about:

> Abstration is a technique for arranging code into different levels of complexity. At the highest level is code that can be easily understood by a person. At the _very_ lowest level is binary code (e.g. 00101010110) that can be understood by computer hardware. There are many levels of abstraction in between, so programmers don't have to interact with details of the system that are not revelant to them (such as how the programming language works, or how memory gets allocated).

js-core-2/lesson5.md
```diff
@@ -33,21 +32,107 @@ Find all the bad design issues with this function
...
+
+> What are abstractions?
+
+In software engineering and computer science, abstraction is a technique for arranging complexity of computer systems. It works by establishing a level of complexity on which a person interacts with the system, suppressing the more complex details below the current level. 
+
+> Abstractions help us with managing complexity, and achieve better reusability.
+
+So far, in the solutions we wrote, we think in terms of steps: DoStep1, DoStep2, DoStep3 until we solve a problem. We're thinking in the **Verbs** that constitute a solution. Our abstractions are **functions** that *do* things to reach a solution. We can use a function without knowing how it works (think about `console.log()`) - the abstraction of that function hides the complexity of how `console.log` actually works.
+
+Now we will step into a new level of abstractions - Objects.
```
djgrant, line 47: Maybe introduce the term Object Oriented Programming and the acroym here.

js-core-2/lesson5.md
```diff
@@ -33,21 +32,107 @@ Find all the bad design issues with this function
...
+
+> The basic idea of OOP is that we use objects to model real world things that we want to represent inside our programs, and/or provide a simple way to access functionality that would otherwise be hard or impossible to make use of.
```
djgrant, line 49: Not sure objects always model "real world things". Maybe just "things"?

js-core-2/lesson5.md
```diff
@@ -33,21 +32,107 @@ Find all the bad design issues with this function
...
+
+## OO overview
+
+### Defining an object template
+
+![](https://mdn.mozillademos.org/files/13889/person-diagram.png)
```
djgrant, line 55: I think a comment like, "In OOP we create templates for objects that we will create. These are often called *classes*."

js-core-2/lesson5.md
```diff
@@ -33,21 +32,107 @@ Find all the bad design issues with this function
...
+
+### Creating actual objects
+
+![](https://mdn.mozillademos.org/files/15163/MDN-Graphics-instantiation-2-fixed.png)
```
djgrant, line 59: And: "From a class we can create (or *instantiate*) an object. This process is called *instantiation*."

js-core-2/lesson5.md
```diff
@@ -33,21 +32,107 @@ Find all the bad design issues with this function
...
+
+### Specialist classes
```
djgrant, line 61: I'm not sure how helpful this is.

js-core-2/lesson5.md
```diff
@@ -33,21 +32,107 @@ Find all the bad design issues with this function
...
+  bio: function() {
+    alert(this.name[0] + ' ' + this.name[1] + ' is ' + this.age + ' years old. He likes ' + this.interests[0] + ' and ' + this.interests[1] + '.');
+  },
+  greeting: function() {
+    alert('Hi! I\'m ' + this.name[0] + '.');
+  }
+};
+```
+
+So what is going on here? Well, an object is made up of multiple members, each of which have a name (e.g. name and age above), and a value (e.g. ['Bob', 'Smith'] and 32). Each name/value pair must be separated by a comma, and the name and value in each case are separated by a colon. The syntax always follows this pattern:
```
djgrant, line 93: Can we use the term `property` instead of `member`?

djgrant: From MDN:

> It's typical when speaking of an object's properties to make a distinction between properties and methods. However, the property/method distinction is little more than a convention. A method is simply a property that can be called, for example if it has a reference to a Function instance as its value.

js-core-2/lesson5.md
```diff
@@ -33,21 +32,107 @@ Find all the bad design issues with this function
...
+
+The value of an object member can be pretty much anything — in our person object we've got a string, a number, two arrays, and two functions. The first four items are `data items`, and are referred to as the object's properties. The last two items are functions that allow the object to do something with that data, and are referred to as the object's `methods`.
+
+> Methods are functions that allow the object to do something with its data.
+
+> The *this* keyword refers to the current object the code is being written inside — so in this case this is equivalent to person. So why not just write person instead?
```
djgrant, line 99: `this` is not a reference to the object, it's the context when the method is called. Assuming the method is called as `person.bio()` then `this` will be `person`, but only because the function `bio` was called from `person`.

Here's an example where `this` does not refer to the object:

```
var bio = person.bio;
bio(); //TypeError: this.interests is undefined (because `this` points to `window`)
```

djgrant: I'm not sure how deep you go into this with beginners but a lot of developers are misled by how `this` works so I think we need to be careful.

js-core-2/lesson5.md
```diff
@@ -33,21 +32,107 @@ Find all the bad design issues with this function
...
+
+So what is going on here? Well, an object is made up of multiple members, each of which have a name (e.g. name and age above), and a value (e.g. ['Bob', 'Smith'] and 32). Each name/value pair must be separated by a comma, and the name and value in each case are separated by a colon. The syntax always follows this pattern:
+
+The value of an object member can be pretty much anything — in our person object we've got a string, a number, two arrays, and two functions. The first four items are `data items`, and are referred to as the object's properties. The last two items are functions that allow the object to do something with that data, and are referred to as the object's `methods`.
+
+> Methods are functions that allow the object to do something with its data.
+
+> The *this* keyword refers to the current object the code is being written inside — so in this case this is equivalent to person. So why not just write person instead? 
+
+An object like this is referred to as an object literal — we've **literally** written out the object contents as we've come to create it. This is in contrast to objects instantiated from classes, which we'll look at next.
```
djgrant, line 101: Probably more accurate to say "An object declaration like this..."

js-core-2/lesson5.md
```diff
@@ -33,21 +32,107 @@ Find all the bad design issues with this function
...
+function Person(name) {
+  this.name = name;
+  this.greeting = function() {
+    alert('Hi! I am ' + this.name + '.');
+  };
+}
+```
+
+This is called a **constructor function**.
+> The constructor function is JavaScript's version of a class. You'll notice that it has all the features you'd expect in a function, although it doesn't return anything or explicitly create an object — it basically just defines properties and methods. You'll see the this keyword being used here as well — it is basically saying that whenever one of these object instances is created, the object's name property will be equal to the name value passed to the constructor call, and the greeting() method will use the name value passed to the constructor call too.
```
djgrant, line 121: Second half of this paragraph could be clearer. How about:

> ...it doesn't return anything or explicitly create an object — rather, it defines properties and methods of a new object that gets created and returned when the function is *instantiated*. That object is called an *instance* of `Person`.

```js
var sally = new Person('Sally');
console.log(sally); // { name: 'Sally': greeting: <Function> }
console.log(sally instanceof Person); // true
```

> You'll see the `this` keyword is being used here as well. What exactly is it doing? When the constructor function is called, `this` is a reference to the object it creates (its instance). In our example we add a property (`name`) and a method (`greeting`) to `this` so whenever Person is instantiated it returns an object with a `name` property and a `greeting` method. 

> You'll also notice that the constructor function take a parameter, `name` (which is getting passed to `Person` when it is instantiated in the example above). When `Person` is instantiated the `name` parameter is assigned to the `name` property of the instance (`this.name = name`) as well as being used to configure the `greeting` method.

js-core-2/lesson5.md
kabaros: yup, I copy/pasted that explanation but it's not accurate at all. We might come back to *this* again, but for now, it's enough to say it refers to the context, which is person in this case.

PR #92, opened 2017-07-02
Syllabus rewrite

Amwam:
Includes a merge of the content from both scotland and london

Old lessons left in place, just in case


kkarimi (approved):
Nice work

PR #97, opened 2017-07-11
Scotland - Scott lesson 7 (JS Core 2)

scott-ad-riley:
Updated lesson notes for Saturday 15th lesson 7 - specifically around object methods and working with objects. Added new set of exercises and examples/mentor notes for updated sections.

### DIFF
--- a/js-core-2/lesson7.md
+++ b/js-core-2/lesson7.md
@@ -19,7 +19,10 @@ We have been dealing mainly with two types of data structures so far - *Arrays* 
 
 When we are you solving a problem, one of your main decisions as a developer is to choose the data structures you will use to solve it. So when do you use an Array and when do you use an Object.
 
-> Let's talk about the differences between Arrays and Objects - when can you use or the other
+> "Bad programmers worry about the code. Good programmers worry about data structures and their relationships."
+> Linus Torvalds
+
+> Let's talk about the differences between Arrays and Objects - when can you use or the other. As a general rule of thumb:
 
 1. Does the order of data matter? Then use Arrays.
 2. Can the data be organised by a label? Then use Objects.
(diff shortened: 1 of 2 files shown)


js-core-2/lesson7.md
```diff
@@ -32,59 +32,60 @@ When we are you solving a problem, one of your main decisions as a developer is
...
     interests : ["javascript", "react"]
 };
 
 printProperties(student);
 // Output in this case should be name, age, interests
 ```
 
+Exercise
+Write a function called hasProperty that takes an object and a property
+the function should return true if the property exists, false if it doesn't
```
rarmatei: The formatting for this is a bit broken: exercise should stand out a bit more. And a new sentence should start after "and a property. The function should ..."

js-core-2/lesson7.md
```diff
@@ -32,59 +32,60 @@ When we are you solving a problem, one of your main decisions as a developer is
...
+    age: "21",
     interests : ["javascript", "react"]
 };
 
 hasProperty(student, 'age'); // should return true
 hasProperty(student, 'job'); // should return false
 ```
 
-> Exercise: compare that to how you find an element in array?
+Exercise: compare that to how you find an element in array?
```
rarmatei: Maybe try and format these exercises to stand out a bit from the rest of the theory, see below:

image

js-core-2/lesson7.md
```diff
@@ -32,59 +32,60 @@ When we are you solving a problem, one of your main decisions as a developer is
...
 
+Exercise
+Write a function called printObject that takes an object and iterates through
+all its properties and prints a string formatted property: value
+Bonus point if you can format the list of interests properly
```
rarmatei: "Bonus point" doesn't start on a new line

js-core-2/lesson7.md
```diff
@@ -98,6 +99,48 @@ printArray(student);
 // "name is Mohamed, age is 22, interests are ["javascript", "c#"]"
 ```
 
+## More advanced working with objects
```
rarmatei: for some reason GitHub is not formatting this for me properly 

image

js-core-2/lesson7.md
```diff
@@ -32,59 +32,60 @@ When we are you solving a problem, one of your main decisions as a developer is
```
rarmatei, line 35: I like the exercises. Could we maybe add a bit more theory to this? I.e. introduce some of the methods they can use `hasOwnProperty` etc. or how they can access properties dynamically by string value: `person["name"]`

scott-ad-riley: These were already in the notes - I've added the wallet exercises underneath but I've left these in the lessons as a sort of recap. I'll look into rewriting them so they're more in-depth with how you can actually use objects.

scott-ad-riley: `hasOwnProperty` is tough to cover without going into more detail around prototypes (which I understand are later in the lesson? or have they covered them already to an extent?). I've added an exercise using it but i'm not sure if it's really adding much value...

I'll have a think about the dynamic property stuff tonight - I'm hesitant to set anything in stone though. I'd rather try to adapt the lesson as they go.

PR #103, opened 2017-07-20
Node.js class 1

Michael-Antczak:


node-db/lesson1.md
```diff
@@ -9,8 +9,132 @@
...
+   
+      res.end('Invalid Request!');
+      
+   }
+});
+
+server.listen(5000); 	
+
+console.log('Node.js web server at port 5000 is running..')
+```
```
kabaros: it looks great .. could you just use const and let for the http example to be consistent with the rest .. also just have the two routes example (no need for single route imo).

Then maybe have an exercise for them to add a /mentor route

kabaros (changes requested):

kabaros (approved):

PR #105, opened 2017-07-21
Work on Node.js workshop

Michael-Antczak:


node-db/lesson1.md
```diff
@@ -581,3 +615,9 @@ Take a look at the following links to learn more about Node.js.
 More about HTTP:
 [Tutsplus tutorial](https://code.tutsplus.com/tutorials/http-headers-for-dummies--net-8039)
 
+Learn more about JSON:
+* http://www.json.org
```
kabaros, line 619: that json.org resource is very confusing

node-db/lesson1.md
```diff
@@ -135,6 +135,42 @@ At the endo of the wizard, you should see a new file called `package.json` in yo
 
 Here is an example `package.json` file for a project called [Passport](https://github.com/jaredhanson/passport/blob/master/package.json).
 
+# What is JSON?
```
kabaros, line 138: this is a better explanation but I think it doesn't need to come that early .. the location where it was fits better imo

Michael-Antczak: I think the topic should be more or less explained at a point when it is introduced, and this is when we talk about package.json

node-db/lesson1.md
Michael-Antczak: Ok, so remove it then.

node-db/lesson1.md
kabaros: ok

kabaros (approved):

PR #109, opened 2017-07-26
Lesson 1 for node.js module from the scotland branch

alexfdz:
#8 cherry-picked lesson 1 for node.js module from the scotland branch


node-db/lesson1.md
```diff
@@ -1,32 +1,673 @@
...
-Please complete the two tasks in the ["before you start"](https://github.com/node-girls/express-workshop#before-you-start) section before you arrive.
+Let's take a website for example.  A website is just a collection of HTML and CSS files, images, maybe some javascript files. When you type a website address in your browser's address bar, the browser (client) sends a **request** to the server that lives at that address. The browser asks the server to give it the files it needs to display the website properly.
 
-The workshop will start at **10:30 AM** in the usual location. Please make sure to arrive on time!
+![Server flow](https://files.gitter.im/heron2014/FiiK/server.png)
 
-### Homework
+# What is Node.js?
+
+  > [Node.js®](https://nodejs.org/en/) is a **JavaScript runtime** built on Chrome's **V8 JavaScript engine**. Node.js uses an **event-driven**, **non-blocking** I/O model that makes it lightweight and efficient.
```
leggsimon, line 26: And in plain english? Might be quite nice to explain "runtime", "event-driven" and "non-blocking"?

Also probably no need for the ® 😉

node-db/lesson1.md
```diff
@@ -1,32 +1,673 @@
...
+	
+	console.log("New request to main page at " + Date())
+        
+        // set response header
+        res.writeHead(200, { 'Content-Type': 'text/html' }); 
+        
+        // set response content    
+        res.write('<html><body><h1>This is home Page.</h1></body></html>');
+		res.write('<h2>The time is: ' + Date() + '</h2>');
+        res.end();
```
leggsimon, line 67: some indentation issues in this block

node-db/lesson1.md
```diff
@@ -1,32 +1,673 @@
...
+- some of the biggest companies use Node.js in production: Netflix, Walmart, IBM, etc.
+- JavaScript everywhere (used to be PHP, Python, JavaScript, MySQL, Apache, now JavaScript full stack)
+
+# A simple Node.js server
+
+**Simple server**
+```javascript
+const http = require('http'); 	
+
+const server = http.createServer(function (req, res) {
```
leggsimon, line 42: I don't feel massively strongly about this but if we're doing ES6, maybe we should use arrow functions? Happy to not though

node-db/lesson1.md
```diff
@@ -1,32 +1,673 @@
...
+
+![starting npm init](https://raw.githubusercontent.com/node-girls/workshop-cms/master/readme-images/step1-npm-init01.png)
+
+It will ask you a bunch of questions.
+
+> You can skip most of the questions but change the `entry point` from `(index.js)` to `server.js`.
+
+> The wizard asks you for the following information: `name`, `version`, `description`, `main`, `test`, `repository`, `keywords`, `author`, `license` - do you understand all of them?
+
+At the endo of the wizard, you should see a new file called `package.json` in your project's folder.
```
leggsimon:
```diff
- endo
+ end
```

node-db/lesson1.md
```diff
@@ -1,32 +1,673 @@
...
 * Complete the nodeschool.io 'learnyounode' tutorial https://github.com/workshopper/learnyounode
-* Research the following topics: blocking vs. non-blocking; RESTful APIs; templating engines; HTTP codes; asynchronous programming. Google them! We will be asking questions in next week's session.
 
 **Important:** When you get stuck during the homework you should follow the following "escalation policy".
 * First, search for answers on Google and Stack Overflow. If you have an error message, copy and paste it into Google.
-* When you are stuck for longer than 20 minutes - no longer! - write a question to your fellow students in the #class1 channel on slack. Take a while to think carefully about the question. This is an important skill. There is a great guide to writing good questions here: http://stackoverflow.com/help/how-to-ask
+* When you are stuck for longer than 20 minutes - no longer! - write a question to your fellow students in the #scotland-class1 channel on slack. Take a while to think carefully about the question. This is an important skill. There is a great guide to writing good questions here: http://stackoverflow.com/help/how-to-ask
```
leggsimon: `#scotland-class1`? or `#london-class1`

node-db/lesson1.md
kabaros: yup .. we asked them to use arrow functions when they try the code and do exercises, but we left optional as few are still not comfortable with it (although I think the London has spent more time with ES6 than here)

node-db/lesson1.md
peterthompson: I think it would be awkward to drop those terms without a full explanation of each. If you want to explain what Node is without them you could write something like this...

Your Chrome browser contains a JavaScript runtime (or engine) called V8 and a Web API for performing asynchronous operations such as XHR or `setTimeout`. Node is a copy of that engine but with it's own Web API for performing IO operations.

...it may be worth then introducing what is meant by a single-threaded non-blocking IO as part of a conversation about how a JavaScript runtime is built. It is a really important concept that underpins how the language works.

node-db/lesson1.md
peterthompson: I'd push to drop in ES6 syntax, it's going to serve them best getting work.

node-db/lesson1.md
```diff
@@ -1,32 +1,673 @@
...
+
+**Simple server**
+```javascript
+const http = require('http'); 	
+
+const server = http.createServer(function (req, res) {
+	res.end("Hello World!")
+});
+
+server.listen(5000);
```
peterthompson, line 46: I'd stick the console log as a callback to the listen function...

```javascript
server.listen(5000, () => {
  console.log('Node.js web server at port 5000 is running...');
});
```

Otherwise on error it will still say that the server is listening to stdout.

node-db/lesson1.md
```diff
@@ -1,32 +1,673 @@
...
+
+> The wizard asks you for the following information: `name`, `version`, `description`, `main`, `test`, `repository`, `keywords`, `author`, `license` - do you understand all of them?
+
+At the endo of the wizard, you should see a new file called `package.json` in your project's folder.
+
+Here is an example `package.json` file for a project called [Passport](https://github.com/jaredhanson/passport/blob/master/package.json).
+
+# What is JSON?
+
+JSON is a type of file for structuring data in a readable way. It is also a really popular format for sending data across the web.
```
peterthompson: I'd mention what the acronym stands for, JavaScript Object Notation.

Also I'd add...
```diff
-a readable way
+a human-readable way
```

node-db/lesson1.md
```diff
@@ -1,32 +1,673 @@
...
+  "children": [],
+  "spouse": null
+}
+```
+
+# Step 2 - Installing Express
+
+Before we write any code, you'll need to install the Express library. We're going to use the **[Node Package Manager (npm)](https://www.npmjs.com/)** to download it using the **`npm install`** command.
+
+NPM is the place to go to download other Node code written by other people.  There are thousands of open-source, 3rd-party Node modules (also known as "packages") by other people that you can download and use in your own projects.
```
peterthompson, line 178: I'd make a very clear distinction between a registry and a client. npmjs.org is the registry where packages can be found. `npm` is a client for downloading and managing those packages. You could go on to mention `yarn` is another example of an npm client that you may encounter.

node-db/lesson1.md
```diff
@@ -1,32 +1,673 @@
...
+
+We already installed Express in Step 2, but we need to make sure it is included in this file specifically so we can make use of its methods. In Node.js, when you want to access the functionality of a library or module in another file, you `require` it.
+
+To import Express, write the following inside `server.js`:
+
+```js
+const express = require('express');
+```
+
+> We will try and get used to ES6 syntax - so use `const` and `let` instead of `var`.
```
peterthompson, line 209: Worth a discussion on what `const` and `let` are with regard to scope?

node-db/lesson1.md
Michael-Antczak: The exercise here was to get them used to reading specs and docs. After reading the sentence we went word by word and explained in plain English. So it wasnt that scary after all.

node-db/lesson1.md
Michael-Antczak: The other point is that this way it was easier to explain passing a function as an argument using a similar syntax to what they know.

peterthompson (approved):

node-db/lesson1.md
alexfdz: @leggsimon @peterthompson do you agree with @Michael-Antczak?

node-db/lesson1.md
alexfdz: @peterthompson I'm moving the code examples to a different repo so should be easier to them to clone, check the code and execute it.

https://github.com/CodeYourFuture/node-workshop

Could you push there your changes in the syntax?

@kabaros from you experience with the module in Scotland, do you think that is a good idea to move the examples to the repo?

node-db/lesson1.md
alexfdz: Fixed in the node-workshop repository

node-db/lesson1.md
kabaros: we haven't tried that before @alexfdz but it seems like something reasonable to do in this case, maybe just prepare some small changes that they do to the code. I would give this approach a try!

node-db/lesson1.md
alexfdz: Could you push it to https://github.com/CodeYourFuture/node-workshop?

node-db/lesson1.md
alexfdz: The content is now in the london branch, feel free to add a bit of magic to this description

node-db/lesson1.md
alexfdz: Fixed in the node-workshop repository
https://github.com/CodeYourFuture/node-workshop/blob/master/scripts/simple_server.js
https://github.com/CodeYourFuture/node-workshop/blob/master/scripts/server_with_routes.js

node-db/lesson1.md
alexfdz: I have created #112 to deal with this conversation. @peterthompson I don't know why I cannot assign it to you.

PR #110, opened 2017-07-27
Node lesson 2

kabaros:
structure and initial content


djgrant (approved):
Added a few edits in commit. Feel free to revert anything I misunderstood :_)

node-db/lesson2.md
```diff
@@ -1,22 +1,392 @@
...
 
-[Link](/node-recap.md)
+> Let's run the server and check it in the browser. This is what you should see on your browser:
+![](assets/blog-screenshot.png)
 
-### Practical exercises: The Status Code API and The Status Code Website.
+* The theme used for this website is based on Bootstrap. You can get more open source Bootsrap-based themes from [StartBootstrap](https://startbootstrap.com/template-categories/all/)
 
-Part 1. [Build a Node status code API](https://github.com/Code-Your-Future/node-status-code-api)
+> **Exercise**: Notice how the link to **Contact me** doesn't return a page - Let's add an enpoint to return a simple string **this is a contact page**.
```
rarmatei: In the homework they've already been asked to return a page, so they should know how to do that. Should we just ask them to return a full page rather than just a string?

node-db/lesson2.md
```diff
@@ -1,22 +1,392 @@
...
 
-Part 2. [Build a Node status code website](https://github.com/Code-Your-Future/node-status-code-website)
-0Looking
+# Templating
+> **Exercise**: Let's change the text for the link, **Contact**, to **Contact Information**.
+
+    - How many files do you have to change? Wouldn't be nice if we could change one file and that reflects in all pages.
```
rarmatei: Possible broken formatting: don't know if you intended this to be in the same block as "Exercises"
image

node-db/lesson2.md
```diff
@@ -1,22 +1,390 @@
...
+> **Exercise**: At Let's change the text for the link for **Contact** from **Contact me** to **Contact Information**.
+
+    - How many files do you have to change? Wouldn't be nice if we could change one file and that reflects in all pages.
+
+At the moment, we're just serving static HTML files from the *public* folder. **NodeJS** and **ExpressJS** allows us to build dynamic web pages.
+
+> A server-side dynamic web page is a web page whose construction is controlled by an application server processing server-side scripts. In server-side scripting, parameters determine how the assembly of every new web page proceeds, including the setting up of more client-side processing.
+
+## Template Engines
+A template engine enables you to use static template files in your application. At **runtime**, the template engine replaces variables in a template file with actual values, and transforms the template into an HTML file sent to the client. This approach makes it easier to design an HTML page, as it reduces duplicate code (DRY).
```
rarmatei: Worth explaining what `static template files` are? They've just been exposed to HTML files up to now.

node-db/lesson2.md
kabaros: this is just part of the recap .. a warmup :)

node-db/lesson2.md
kabaros: yeah I was thinking of explaining in words and examples in the class, cause I found it really hard to put it in words here (this is the explanation from express docs) ... what would you say to explain it better here? @rarmatei

node-db/lesson2.md
```diff
@@ -1,22 +1,390 @@
...
+}
+
+delay(logRandom);
+```
+
+**Key points**
+* Functions can be passed as arguments to other other function
+* Functions that are passed as arguments are called **callbacks**
+* Callbacks may be executed immediately or later
+* If callback is executed later, its return value is lost
```
rarmatei: Hmm, I might be missing the point, but is this last rule necessary? Why is the return value lost?

Observables are an example of a construct that asynchronously uses the return value of callbacks.

node-db/lesson2.md
```diff
@@ -1,22 +1,390 @@
...
+```
+
+## Setup an error handler
+You define error-handling middleware in the same way as other middleware, except with four arguments instead of three; specifically with the signature (err, req, res, next):
+
+```javascript
+app.use(function (err, req, res, next) {
+  console.error(err.stack)
+  res.status(500).send('Something broke!')
+})
```
rarmatei, line 268: Does it matter where we add the middleware in our chain? Does it work the same way added in the beginning or at the end?

node-db/lesson2.md
```diff
@@ -1,22 +1,390 @@
...
+var myLogger = function (req, res, next) {
+  console.log('LOGGED')
+  next()
+}
+
+app.use(myLogger)
+```
+
+> Exercise: Follow the tutorial for writing `requestTime` at https://expressjs.com/en/guide/writing-middleware.html and add it to your own application
+
```
rarmatei: Is middleware arrangement in the chain important? Should we discuss that at some point?

node-db/lesson2.md
```diff
@@ -1,22 +1,390 @@
...
+Now push your code to heroku `git push heroku master`. The push will run few commands from Heroku, then you should see a url similar to `https://some-random-name-XXXX.herokuapp.com` - go to the Url and if all goes well, your app should be up and running.
+
+To read more about Heroku and deploying Node Apps to Heroku, check
+https://devcenter.heroku.com/articles/git and https://devcenter.heroku.com/articles/getting-started-with-nodejs#set-up
+
+# Homework
+- Deploy to Heroku if you haven't yet
+- Implement the Admin page. Write a posts endpoint that you can hit and that should save to the JSON file (use the helper functions we added)
+- Secure the Admin page so that it's only visible if a certain query parameter is provided
```
rarmatei: Have we covered query parameters?

node-db/lesson2.md
```diff
@@ -1,22 +1,394 @@
...
+To read more about Heroku and deploying Node Apps to Heroku, check
+https://devcenter.heroku.com/articles/git and https://devcenter.heroku.com/articles/getting-started-with-nodejs#set-up
+
+# Homework
+- Deploy to Heroku if you haven't yet
+- Implement the Admin page. 
+- Write a posts endpoint that you can hit and that should save to the JSON file (use the helper functions we added)
+- Secure the Admin page so that it's only visible if a certain query parameter is provided
+    - Can you go a bit further with adding proper security? Research the internet for solutions in **Express.js**
+- Add a route `posts/:postid` that the displays a specific post
```
rarmatei: `that displays`

node-db/lesson2.md
```diff
@@ -1,22 +1,394 @@
...
+- Consume a posts API built by another colleague (and deployed to Heroku) to display their latest blog posts on
```
rarmatei: blog posts on...?

rarmatei:
Should we move Heroku up in the lesson? It's needed for the homework, but we might not get to it since this lesson is a bit heavy on new content.

node-db/lesson2.md
kabaros: yeah for this one, it doesn't matter. It's about having four parameters, it's a special middleware.

kabaros:
@rarmatei yeah, we will do heroku at some point in the middle of the lesson. Just thought to leave at the end of this file, to not break the flow as it's a separate topic.

node-db/lesson2.md
rarmatei: "A template file looks like a normal HTML file with tags and attributes, but you can also apply a special template programming language to it, for more advanced manipulation, as we'll see in the following examples." Up to you, we don't need to put in the lesson, like you said, we can explain in the class.

rarmatei (approved):

PR #119, opened 2017-08-16
Work on Node class 3

Michael-Antczak:

### DIFF
--- a/node-db/lesson3.md
+++ b/node-db/lesson3.md
@@ -127,7 +127,7 @@ Let's look at the [AWS Database](https://aws.amazon.com/products/databases/) sec
 
 SQL vs. NoSQL
 
-## Store posts to DynamoDB
+## Using DynamoDB with our Express app
 
 What is [AWS DynamoDB](https://aws.amazon.com/dynamodb/)? What is the price? Where to find docs?   
 Watch the video [Introduction to Amazon DynamoDB](https://aws.amazon.com/dynamodb/getting-started/)
@@ -193,6 +193,8 @@ We are going to use AWS SDK for this.
         }
     });
     ```
+
+Discussion: do you think this is a good way to read the posts? If not, why not? How can it be done differently? 
 
 # Express Generator
 
--- a/node-db/lesson3.md
+++ b/node-db/lesson3.md
@@ -127,7 +127,7 @@ Let's look at the [AWS Database](https://aws.amazon.com/products/databases/) sec
 
 SQL vs. NoSQL
 
-## Using DynamoDB with our Express app
+## Using [DynamoDB with our Express app](http://docs.aws.amazon.com/amazondynamodb/latest/gettingstartedguide/GettingStarted.NodeJs.html)
 
 What is [AWS DynamoDB](https://aws.amazon.com/dynamodb/)? What is the price? Where to find docs?   
 Watch the video [Introduction to Amazon DynamoDB](https://aws.amazon.com/dynamodb/getting-started/)
@@ -149,6 +149,14 @@ We are going to use AWS SDK for this.
 
 1. run `npm install --save aws-sdk`
 1. [load credentials](http://docs.aws.amazon.com/sdk-for-javascript/v2/developer-guide/loading-node-credentials-json-file.html) - create in the root of the project a file called `config.json`. We will load are keys from there
+    ```javascript
+    {
+        "accessKeyId": "YOUR_ACCESS_KEY",
+        "secretAccessKey": "YOUR_SECRET_KEY",
+        "region": "eu-west-1",
+        "endpoint": "https://dynamodb.eu-west-1.amazonaws.com"
+    }
+    ```
 1. change `us-east-1` to `eu-east-1` and add the access and secret keys
 1. add AWS to the project
     ```javascript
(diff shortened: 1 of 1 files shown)


docs/guides/marking-guide.md
```diff
@@ -0,0 +1,210 @@
...
+sidebar_label: Marking Guide
+---
+
+This Rubric should be used to guide your assessment of Students Homework at CodeYourFuture
+
+## Overview
+
+### Completeness (/10)
+
+- The code achieves all of the tasks in the coursework
```
illicitonion: Again with the homework/coursework distinction - let's aim for one term if we can?

docs/guides/marking-guide.md
```diff
@@ -0,0 +1,210 @@
...
+- There is no dead code in the coursework [[read more]](#dead-code)
+- There is no extra or debug logging [[read more]](#debug-logging)
+
+## Explanations
+
+### Edge Cases
+
+Edge cases are when code does something you're not expecting and you didn't account for in your code.
+
+For example, using a number in a `Name` field of a form might cause problems since peoples names can't be numbers.
```
illicitonion:
```suggestion
Some edge cases are valid data, and some are not.

Some examples of valid edge case are that some names contain punctuation, or some people may only have a single name, which your code needs to be able to handle. An example of an invalid edge case is if someone doesn't enter a name at all, which your code should reject - if your code doesn't notice the problem where the user provides their name, other code you wrote may make assumptions which are incorrect.

See [Falsehoods Programmers Believe About Names](https://www.kalzumeus.com/2010/06/17/falsehoods-programmers-believe-about-names/) for some interesting examples.
```

docs/guides/marking-guide.md
```diff
@@ -0,0 +1,210 @@
...
+console.log("Does this work");
+console.log("1");
+console.log(array[1].object.name);
+```
+
+should be removed.
+
+### Error Responses
+
+Not just the error messages, every response that is returned by the server must be properly handled. It should have necessary headers, response messages, error codes and any other necessary details attached with it in required format. All possible scenarios are tested to avoid deadlocks, timeouts, etc.
```
illicitonion:
```suggestion
Not just the error messages, every response that is returned by the server must be properly handled. It should have necessary headers, response messages, error codes and any other necessary details attached with it in required format. All possible scenarios are tested to avoid timeouts, unresponsive UI, etc.
```

I feel like deadlocks are a more advanced topic than we cover :)

docs/guides/marking-guide.md
```diff
@@ -0,0 +1,210 @@
...
+
+Each variable is named so it is clear which variable is being used at any time. It is important to use meaningful names for variables:
+
+For example,
+
+```js
+let pocketMoney = 20;
+```
+
+means that the variable ‘pocketMoney’ is being used to store how much pocket money you have. Right now you have £20.
```
illicitonion:
```suggestion
means that the variable `pocketMoney` is being used to store how much pocket money you have. Right now you have £20.
```

docs/guides/marking-guide.md
```diff
@@ -0,0 +1,210 @@
...
+
+The name given to each variable is up to the programmer, but ideally a variable name should have meaning, ie it should reflect the value that it is holding.
```
illicitonion:
```suggestion
The name given to each variable is up to the programmer, but ideally a variable name should have meaning you can understand without needing to read any other code, i.e. it should reflect the value that it is holding.
```

docs/guides/marking-guide.md
```diff
@@ -0,0 +1,210 @@
...
+function doWork() {}
+```
+
+A good function name might be
+
+```js
+function removeCaptialLetters(){)
+```
+
+Sometimes if you can't come up with a good function name it's an indication that the function doesn't have a nice, crisp focus and needs to be refactored. If it's a class method, perhaps the class needs refactoring too.
```
illicitonion:
```suggestion
Sometimes if you can't come up with a good function name it's an indication that the function doesn't have a nice, crisp focus and needs to be refactored.
```

We don't teach classes :)

Also, I'm not sure we ever teach/explain the term refactoring?

docs/guides/marking-guide.md
```diff
@@ -0,0 +1,210 @@
...
+means that the variable ‘pocketMoney’ is being used to store how much pocket money you have. Right now you have £20.
+
+The name given to each variable is up to the programmer, but ideally a variable name should have meaning, ie it should reflect the value that it is holding.
+
+#### More Reading
+
+- [Good Variable Names](https://spin.atomicobject.com/2017/11/01/good-variable-names/)
+- [Programming Basics - Naming Variables](https://www.bbc.co.uk/bitesize/guides/zwmbgk7/revision/3)
+
+### More Optimal Code
```
illicitonion: I'm not sure about this section for a couple of reasons

1. Optimal in my mind means efficient, not readable
2. Our learners should favour readability over efficiency most of the time
3. We don't really teach anything about efficiency.

I feel like I'd rather frame this as something like:

> ### More readable code
> 
> The most important thing when writing code is to make it easy to read and understand, and as simple as possible.
> 
> [contrasting examples of `filter` and `for` go here - ideally two examples, one showing a case where `filter` is simpler, one showing a case where a `for` loop is simpler]

(And update the bullet which links here to talk about readability not optimality)

If we want to talk efficiency, we can do that too, but I suspect we don't.

docs/guides/marking-guide.md
```diff
@@ -0,0 +1,210 @@
...
+Hard coded values with no explanation make code hard to understand and read in the long term. For example
+
+```js
+getData("https://www.api.com/retrieve");
+calculateScreenSize(1800);
+```
+
+Will be hard to understand when you codebase grows.
+
+Instead us constants to store your hard coded values. For example
```
illicitonion:
```suggestion
Instead us constants to store your hard coded values - this means the value has a name, which can be [usefully descriptive](#good-variable-names). For example
```

docs/guides/marking-guide.md
```diff
@@ -0,0 +1,210 @@
...
+
+Validations are used wherever necessary. The code never breaks under any circumstances. Especially under invalid inputs that come from the user end. Regardless of it being a negative, over-sized, invalid format, etc., every input passed should be processed, sanitized before taking it further. Every object is checked for its actual data existence before accessing its properties.
+
+#### More Reading
+
+- [Don't forget edge cases](https://www.geeksforgeeks.org/dont-forget-edge-cases/)
+
+### Small Functions
+
+Writing small functions makes your code easier to understand and easier to work with
```
illicitonion:
```suggestion
Writing small functions makes your code easier to understand and easier to work with. Ideally you should be able to describe what a function does in one sentence, without the word "and".

Good examples: "This function returns the user's favourite author", "This other function Tweets the title of an author's most popular book".

Bad example: "This function finds the user's favourite author and sends a Tweet with the name of that author's most popular book".
```

docs/guides/marking-guide.md
```diff
@@ -0,0 +1,210 @@
...
+
+Duplicated code should be moved into functions so they can be reference from multiple places.
+
+Having lots of duplicated code in your coursework means that it is harder to change or update your code.
+
+#### More Reading
+
+- [Duplication Code](https://dev.to/thinkster/code-smell-duplicated-code-fc1)
+
+### Commented Out Code
```
illicitonion, line 137: Maybe worth plugging git history here?

docs/guides/marking-guide.md
```diff
@@ -0,0 +1,210 @@
...
+#### More Reading
+
+- [Please don't comment out code](https://kentcdodds.com/blog/please-dont-commit-commented-out-code)
+- [Putting comments in code: the good, the bad, and the ugly.](https://www.freecodecamp.org/news/code-comments-the-good-the-bad-and-the-ugly-be9cc65fbf83/)
+
+### Dead Code
+
+Dead Code is code that is in your coursework but can't be reached or run by your program.
+
+For example, a function that is never called (or invoked) is considered dead code.
```
illicitonion:
```suggestion
For example, a function that is never called is considered dead code.
```

PR #127, opened 2017-09-01
Introduction to SQL

jamesseanwright:
Using existing interactive resources in lieu of custom database and content.


kabaros (approved):

kabaros:
I would think - depending on how the class goes - we could drop altering tables in favor of something like Joins? It seems like a concept that would be more important at that level.

Otherwise sqlbot looks great

jamesseanwright:
@kabaros SQLBolt does cover joins, so we'll be looking at them together once the students have completed the initial `SELECT` statement exercises.

Issue #131, opened 2017-09-11
React module


kabaros:
This is a Proposed skeleton for the React module.

The need for a review is caused by few differences (and lessons learned from cohort 1):
- This round, the students did **Node** before **React** so no need to familiarise them with npm, packages etc...
- The students are already familiar with ES6 syntax
- Lessons learned: We gave React Router as a study subject to encourage the students to learn by their own - that didn't work at all.
- Reviewing the homework and exercises: The model where we have a continuous project that they build and improve during a module worked well. It will be good to have that sense of continuity for their exercises and homework this time around as well.

The module starts on September 30th in Scotland. This is the proposed skeleton of what to cover, and the order. You can have a look at it here https://codeyourfuture.github.io/syllabus-scotland/react/lesson1.html . Please ignore the contents of the lesson themselves - this is a quick dumb from the previous round (London cohort 1) that we should update with more relevant materials.

Please check the contributing guidelines on how to contribute to this module.

# React 1
- React
  - Why React?
  - What is an SPA?
  - create-react-app
- Components
  - JSX
  - ES6 classes
  - Thinking in React
- JSX Patterns
  - Mapping Arrays
  - Conditionals
  - Embedding code
- props and state

# React 2
- More Props and State
  - Stateful (smart) vs Stateless (dumb) components
- Component life cycle
  - componentDidMount: Making API calls
- Testing React

# React 3
- React Router
- More testing

# React 4
- React Native Workshop

ghost:
If we cover OOP with es6 earlier in the JS modules, we can just nod to it in React. Also we could sum up JSX as ‘a way to embed Javascript into react applications’ and just focus on patterns.

The problem with React is that there's so much tooling involved compared to everything else so I'm thinking we could avoid confusion by now diving deep into JSX or React Router, focusing instead on practical application of the tools.

I also wouldn't focus as much on testing React components during the modules. Plenty of good resources out there on how to test React comps and because we've TDDed all the way through the past modules the behaviour should be engrained.

The discussion of stateless and stateful comps could be changed to presentational / containers and that can be the springboard into redux.

kkarimi:
It needs to be noted that the main feedback we had from last time was that the students really **struggled** with it, and just about got it, though it mostly became clear in the actual project. Better fundamental JS and maybe Handlebars if it was covered this time around may help students understand this module better but more importantly they need to get engaged with a project to really get the react concepts and think in the SPA terms.

[...]

PR #132, opened 2017-09-25
React lesson 1

rarmatei:
If this lesson were a person, it would wear a large T-shirt. It's so large mainly because it contains a lot of code examples.

The aim of this first lesson is to get them interested in React and learn the basics. I wanted to avoid spending too much time just on explaining theoretical concepts, and instead learn them as we build a game. And then strengthen them during future lessons.

All of the topics I'm covering here are in line with @kabaros [Lesson 1 Plan for React](https://github.com/CodeYourFuture/syllabus/issues/131).

I'm 100% confident we'll be able to get through all the React concepts in this lesson, and even if we don't have time to finish the whole game, I can have them finish it as homework. After the Node/DB module, it should be a lot easier for them to connect the puzzle pieces with React, as it's a lot less new APIs/concepts and more capitalising on previous JavaScript/HTML knowledge. I really hope that will be the case.

These are the only two files they'll be creating during the workshop (will be slightly different, but mostly the same):
https://github.com/rarmatei/CYF-ReactGame/blob/master/src/App.js
https://github.com/rarmatei/CYF-ReactGame/blob/master/src/components/button.js

It doesn't seem to be a lot of code. Maybe manageable for an intro class + what's left for homework?

I still have to arrange it a bit, and make it look nicer.


djgrant:
A couple of pre-review thoughts:

- Should we have a prerequisite task to install prettier? This will make it much easier for both us and the students to read each other's code. And save some time pointing out syntax errors.

- Can we add an "Edit on Code Sandbox" button for all our examples? This will allow students to see the result immediately and go play around. See https://codesandbox.io/s/github/rarmatei/CYF-ReactGame -> Share.

react/lesson1.md
```diff
@@ -16,17 +16,72 @@
 
 # React
 
-[React](https://facebook.github.io/react/) is a library by Facebook.
+[React](https://facebook.github.io/react/) is a library by Facebook for building user interfaces. 
+
+Today, the focus is on component based architectures, and React provides a thin layer on top of JavaScript and HTML to allow you to build custom components that can encapsulate behavior and style and can be re-used throughout your application and even exported for others to use.
```
djgrant: Can you break this into smaller sentences or bullets?

djgrant: A definition of 'encapsulation' would be useful.

react/lesson1.md
```diff
@@ -16,17 +16,72 @@
...
+..it generates this: 
+
+<img src="assets/html_input.png" width="150">
+
+An "input" box, that only supports numbers, with a set of up/down arrows for changing the value, and a pre-defined value of 5. So the **element tag** tells the browser which element to render, and the attributes specify the settings for that element and any information we want to pass into it.
 
 ## Why React?
-ToDo
+
+Besides providing the tools to build powerful and well encapsulated components, React also takes care of updating the view to match our data.
```
djgrant: Are students familiar with the term "view" in this context. If we can keep the set of vocabulary we use small that will help. Perhaps "rendered component" would be better.

react/lesson1.md
```diff
@@ -16,17 +16,72 @@
...
+```
+
+And our HTML looks like this:
+
+```html
+<h1 id="title"> </h1>
+<p id="content"> </p>
+```
+
+As we've learned, we can dinamically insert content into our HTML using JavaScript:
```
djgrant: dynamically

react/lesson1.md
```diff
@@ -16,17 +16,72 @@
...
+
+```javascript
+document.querySelector('#title').innerText = post.title;
+document.querySelector('#content').innerText = post.content;
+```
+
+So everytime our `post` variable changes, we have to manually call the above to update our HTML, so the user can actually see the new post.
+
+However, with React, you can just setup your template once:
```
djgrant: "set up"

djgrant: Again, let's stick with react terms, so "component" rather than "template".

react/lesson1.md
```diff
@@ -36,175 +91,431 @@ cd my-app/
 npm start
 ```
 
+# Our first component
 
-# JSX
-- Uses [JSX](https://facebook.github.io/react/docs/introducing-jsx.html)
+Components in React can be defined by just declaring a function that return some JSX.
```
djgrant: A short explanation of what JSX would be good here e.g. JSX is HTML-like code that you can include in JavaScript files.

react/lesson1.md
```diff
@@ -36,175 +91,431 @@ cd my-app/
...
-            website="http://www.refugeesupportnetwork.org"
-        />
-      </div>
-    );
+    return <button> Awesome button! </button>;
   }
 }
+```
+
+The above is the exact same component we defined using the function.
```
djgrant: Would be more accurate to say "equivalent", rather than "exact same". Or that it has the exact same effect.

react/lesson1.md
```diff
@@ -36,175 +91,431 @@ cd my-app/
...
+    return <button> Awesome button! </button>;
   }
 }
+```
+
+The above is the exact same component we defined using the function. 
+
+Child or `presentational` components are essentially smaller customisable to the data that they receive but usually do not have the data in themselves. These are often referred to as `dumb` components - because they do not process the data in themselves:
+
+While function components are a lot simpler to read and write, using ES6 classes allows us to create more powerful, smarter components.
```
djgrant: _functional_ components

react/lesson1.md
```diff
@@ -36,175 +91,431 @@ cd my-app/
...
+
+# JSX
+
+You might have noticed something weird: we were returning some HTML-like code in our components.
+
+That is called JSX, and it's a compromise to allows us to use the HTML we are so familiar inside our JavaScript components.
```
djgrant: _that_ allows

react/lesson1.md
```diff
@@ -36,175 +91,431 @@ cd my-app/
...
+
+You might have noticed something weird: we were returning some HTML-like code in our components.
+
+That is called JSX, and it's a compromise to allows us to use the HTML we are so familiar inside our JavaScript components. 
+
+Since it's not exactly HTML, there are some gotchas, of which you'll need to be aware of:
+
+- event handlers are capitalized
+
+HTML: `<button onclick="myFunction()"/>`
```
djgrant: you wouldn't call `myFunction` in either of these examples

react/lesson1.md
```diff
@@ -36,175 +91,431 @@ cd my-app/
...
+
+3 things to notice:
+
+- the curly brackets after `onChange`: surrounding something by two curly brackets is a way to use JavaScript inside JSX
+- `difficultyChange` is written as an arrow function: this is a good way to avoid issues with the `this` context in React. You can read more about it here: https://facebook.github.io/react/docs/handling-events.html
+- the `event` object: this is a special object passed to all event handlers in React. In our case we can use it to access the target DOM element of that event, and more specifically, its value.
+
+# JSX: Map Arrays to Components
+
+Because works so well with JavaScript, we can even it in for loops and build JS arrays that contain JSX elements.
```
djgrant: Not sure exactly how this sentence is intended to read

react/lesson1.md
```diff
@@ -36,175 +91,431 @@ cd my-app/
...
+  <br/>
+  {this.buildGame()}
+</p>
 ```
 
-- [Props exercise](https://jsfiddle.net/kabaros/1h9taf1e/)
+# State
 
+A component's internal state represents all the information that a components "knows" about itself. 
+State can only exist on `class` components. That is why they're also called `statefull` components.
```
djgrant: _stateful_

react/lesson1.md
```diff
@@ -36,175 +91,431 @@ cd my-app/
...
+
+Let's go over some examples and discuss how we would break everything up into smart and dumb components.
+
+# Event handlers
+
+Time to handle a few user events. We'll add a "difficulty setting" input. For now, whenever the user changes that input, we will `console.log` the result.
+
+```javascript
+
+difficultyChange = (event) => {
```
djgrant: I got a bit confused about what this function would be used for until much later on. Could it be named to something that conveys how the UI will change, for example `changeNumberOfButtons`?

djgrant: And `this.state.difficulty` be changed to `this.state.numberOfButtons` (or `buttonCount` or whatever).

react/lesson1.md
```diff
@@ -36,175 +91,431 @@ cd my-app/
...
+
+# Props
+
+Remember HTML attributes? 
+
+`<input value="Hervin" class="nameInput" />`
+
+Props are the inputs sent from the parent to child component via its attributes. So for our `GameButton` we could send it a prop that would specify what text it should display on the button:
+
+`<GameButton btnText="First button" />`
```
djgrant: I have a preference for not removing vowels from variable names. `btn` is obvious but I wouldn't want students to get the idea that fewer characters = better code.

react/lesson1.md
```diff
@@ -36,175 +91,431 @@ cd my-app/
...
+Then let's give a different css class to our button depending on whether it's selected or not:
 
-State
----
-Props are a Component's configuration, they are received from above and they don't usually change by the component itself.
-State starts with a default value (set in the constructor typically) then it gets changed (mutated) by events of the application (user interactions, form input, API calls etc..)
+```javascript
+import './gameButton.css';
 
+function GameButton(props) {
```
djgrant: Would it be too much at this stage to show how you would refactor this to use a ternary operator?

```js
const GameButton = props => <button className={props.selected ? 'neutral' : 'selected'}>Awesome button!</button>;
```

djgrant (commented):
@rarmatei This is a great lesson. I really like the way it flows and your explanations and really understandable. I've left a few comments for you to consider but they're all pretty minor things. 

The game itself felt a bit to complicated to me on first sight. Do you think there are any ways it could be made simpler or broken into smaller/simpler parts.

react/lesson1.md
```diff
@@ -16,17 +16,72 @@
...
+> **Exercise:** 
+>
+> What does the above HTML remind you of? Any similar technology we studied recently that does something similar?
 
 ## What is an SPA?
-ToDo
+
+With these features in mind, it suddenly becomes a lot easier to build Single Page Applications (SPA). 
+
+SPAs are web pages fit in a single web page, providing a similar experience to desktop applications. All the content a page requires is usually downloaded at once (HTML, JavaScript and CSS), in the first initial request, or loaded dinamically as the user is clicking around the page. There are no reloads as you click from one section to the next.
```
scott-ad-riley: `dinamically` > `dynamically`

scott-ad-riley: Could also be worth a mention of ajax in here as they've covered that before and it's pretty relevant to building an SPA.

react/lesson1.md
NateWr: Or "display" or even "HTML code" since that's what's happening.

react/lesson1.md
```diff
@@ -36,175 +91,431 @@ cd my-app/
...
-      </div>
-    );
+    return <button> Awesome button! </button>;
   }
 }
+```
+
+The above is the exact same component we defined using the function. 
+
+Child or `presentational` components are essentially smaller customisable to the data that they receive but usually do not have the data in themselves. These are often referred to as `dumb` components - because they do not process the data in themselves:
```
scott-ad-riley: the word `process` feels strange here - I feel like they do process it because it comes in as props and comes out as rendered html. It could be more accurate to say they do not modify, touch or create the data (it's a tricky one to reword). A common word is they are not "responsible" for the data - though this could be a concept they've not really come across before.

react/lesson1.md
scott-ad-riley: I'd also argue against the word `compromise` here 😄 

Maybe a "technique"? even "language" works imo

react/lesson1.md
scott-ad-riley: And you'd almost always pull the function off of either `this` or `this.props`.

react/lesson1.md
```diff
@@ -36,175 +91,431 @@ cd my-app/
...
 
+- must always have one child element
+
+Won't work:
+
+```javascript
+return (
+  <p> Hello! </p>
+  <p> This is my blog. </p>
+);
```
scott-ad-riley: I don't know if it's worth mentioning that react 16 will allow you to do this? (if only briefly)

React 16 launch/rc post:https://github.com/facebook/react/issues/10294

react/lesson1.md
```diff
@@ -36,175 +91,431 @@ cd my-app/
...
+
+//inside render()
+<input type="number" onChange={this.difficultyChange}/>
+```
+
+3 things to notice:
+
+- the curly brackets after `onChange`: surrounding something by two curly brackets is a way to use JavaScript inside JSX
+- `difficultyChange` is written as an arrow function: this is a good way to avoid issues with the `this` context in React. You can read more about it here: https://facebook.github.io/react/docs/handling-events.html
+- the `event` object: this is a special object passed to all event handlers in React. In our case we can use it to access the target DOM element of that event, and more specifically, its value.
```
scott-ad-riley: Whilst React does wrap the `event` object that gets passed here, i'd avoid mentioning it. We aren't actually using any of the wrapped behaviour. It behaves exactly the same as if you had written:
```javascript
const input = document.getElementById('input-id')
input.onchange = (event) => { console.log(event.target.value) }
```
i.e. the `event` there is equivalent to the event that comes in to `difficultyChange`

react/lesson1.md
```diff
@@ -36,175 +91,431 @@ cd my-app/
...
-export default Organisation;
+And then call it in our `render()` function:
 
+```javascript
+<p className="App-intro">
+  <input type="number" onChange={this.difficultyChange}/>
+  <br/>
+  {this.buildGame()}
+</p>
 ```
```
scott-ad-riley: Can we include an example with map here? It's mentioned in the heading for the section and it's often much more common to use `this.props.someField.map` over a `for (let i = 0; i < this.props.someField.length; i++)` in React.

Be aware that we're also going to get a warning about not having a `key` on each element here.

react/lesson1.md
```diff
@@ -36,175 +91,431 @@ cd my-app/
...
+constructor() {
+  super();
+  this.state = {
+    difficulty: 10
+  };
+}
+```
 
-```jsx
+Notice the call to `super()`. We always need to call that before referencing the `this` keyword in the constructor. In this case it calls the `constructor()` of the parent class `React.Component`. `super` is not a React function, it's an ES6 keyword used in class inherritence: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/super
```
scott-ad-riley: `inherritence` > `inheritance`

react/lesson1.md
```diff
@@ -36,175 +91,431 @@ cd my-app/
...
-    return orgData.map(function renderData(organisation) {
-        return (
-            <Organisation
-                name= {organisation.name}
-                borough={organisation.borough ? organisation.borough : 'None'}
-                website= {organisation.website}
-            />
-        )
+```javascript
+  difficultyChange = (event) => {
```
scott-ad-riley: Sidenote: this is not currently in ES6 - `create-react-app` will give us [babel-plugin-transform-class-properties](https://www.npmjs.com/package/babel-plugin-transform-class-properties) out of the box.

Not sure if it's worth mentioning - downside is that the alternatives are more complicated to understand (either `.bind(this)` in `render` or in the `constructor`)

react/lesson1.md
scott-ad-riley: yeh for the sake of clarity here - `text` should be fine

react/lesson1.md
```diff
@@ -36,175 +91,431 @@ cd my-app/
...
+.neutral {
+    background: lightgrey;
+}
+
+.selected {
+    background: red;
+}
+```
+
+Now try changing the `selected` attribute on your `GameButton`. It should change color depending on whether it's true or false.
```
scott-ad-riley: I haven't seen where you have mentioned why we need to use `className` over `class`.

(Apologies if i've missed it - but it's likely to be a question that comes up)

react/lesson1.md
```diff
@@ -36,175 +91,431 @@ cd my-app/
...
+
+In your render function, print out the final value:
+
+```javascript
+Score: {
+  this.state.endTime > 0
+    ? this.state.endTime - this.state.startTime
+    : "Not yet started.."
+}
+```
```
scott-ad-riley: can this ultimately get moved to a function outside of render? `displayScore` or something.

react/lesson1.md
```diff
@@ -36,175 +91,431 @@ cd my-app/
...
+    this.setState({
+      selectedButton: randomIndex,
+      startTime: new Date().getTime()
+    })
+  }
+
+  // When the user manages to click on the selected button, display the time passed start and end
+  gameButtonClick = (event) => {
+    const className = event.target.className;
+    if(className.includes('selected')) {
```
scott-ad-riley: This feels like an antipattern - We shouldn't be looking at the state of the DOM to determine whether a clicked element has been selected.

I haven't pulled the repo down locally and ran the app so I can't offer a nice alternative right now but it feels like a closure/bound function that returns this above function, and has the number of the clicked button would be nicer.

scott-ad-riley (commented):
Looks good to me - a few comments in places but I feel like the level of detail is perfect.

This sets up the next set of lessons on ComponentLifecycle, async-ness of `setState` etc. pretty well.

react/lesson1.md
rarmatei: Totally agree. I'll try to pay attention to the terms I'm using in the class as well.

react/lesson1.md
rarmatei: I'll just run the whole thing through an type-checker 😛

react/lesson1.md
rarmatei: Hmm, at this stage they should be very familiar with the concept of template from handlebars, while we've only briefly mentioned components when discussing how they should break down their website. I don't mind going with "component" though.

react/lesson1.md
rarmatei: I might rename JSX here to HTML. JSX is explained a bit later?

react/lesson1.md
rarmatei: Yep, good spot

react/lesson1.md
rarmatei: Will change

react/lesson1.md
rarmatei: I got the term "compromise" from this article: https://medium.freecodecamp.org/all-the-fundamental-react-js-concepts-jammed-into-this-single-medium-article-c83f9b53eac2 (do a CTRL+F for compromise), and I quite liked his explanation.

However, yea, might not be best used in an "intro to React" lesson.

react/lesson1.md
rarmatei: Will change the React example, thanks for pointing out. 

For HTML, that's how you would do it, though probably not recommended, I'll leave as it is as I'm just showing a small difference they ought to be aware of: https://developer.mozilla.org/en-US/docs/Mozilla/Tech/XUL/Attribute/onclick

react/lesson1.md
rarmatei: That makes sense, it's a lot more obvious

react/lesson1.md
rarmatei: Haha, me neither. I think it should've read: "Because JSX works so well with JavaScript, we can even use it in `for` loops and build JavaScript arrays that contain JSX elements."

react/lesson1.md
rarmatei: agree, it's already button, no need for the prefix

react/lesson1.md
rarmatei: I can show that during the class, makes sense. Wanted to as explicit as possible here, though.

react/lesson1.md
rarmatei: Very true, I'll add a mention in there.

react/lesson1.md
rarmatei: Hmm, how about:

Child or `presentational` components are concerned with how things look. They contain very little logic. They are like a function: take some inputs, and return an element (and that's how they're usually written as well, as a function). They don't hold or remember any data: the element that they return is based only on the inputs they received, nothing else.

react/lesson1.md
rarmatei: I might remove this whole section on JSX "gotchas" all-together. Adds a lot of new things for the students to remember, and I don't know if it's necessary to focus on them: most things they will just discover when working with it, and maybe it's going to stick better that way.

react/lesson1.md
rarmatei: Hmm, don't think I've mentioned the "wrapping" of the event object? Sorry if I've missed it 😜

react/lesson1.md
rarmatei: I mentioned it in the JSX section gotchas. Thanks for checking, though!

react/lesson1.md
rarmatei: You are totally right! Although I might leave as it is, going into how the closure works in this case would be a bit too much I think?

react/lesson1.md
rarmatei: It can. Although I mainly wanted to show an example of using a ternary operator in JSX. Don't mind moving it out though if you think it's better?

react/lesson1.md
scott-ad-riley: Yeh - that's great 👍

react/lesson1.md
scott-ad-riley: yep, sounds good.

react/lesson1.md
scott-ad-riley: it came from the phrase "special object passed to all event handlers in React" - it's not actually associated with React at all. I think I could have read a little too much into much into what you've written... but it's a quick thing to fix. The `event` object is passed into any event handler as part of the DOM/Javascript spec.

react/lesson1.md
scott-ad-riley: Yeh - this could be better of as a bit of a refactor to avoid both a closure and this - i'll pull the repo tomorrow and take a look.

react/lesson1.md
scott-ad-riley: I think we've already got a ternary above this, I find multi-line ternaries are quite hard to read when compared against a function with an `if`/`else`

react/lesson1.md
rarmatei: Cool, will change!

react/lesson1.md
rarmatei: I've gotten used to ternaries, especially when on multiple lines, sometimes I find them much easier to read. But the students are definitely more used to a simple if/else, so I'll move it out, as you suggested.

PR #138, opened 2017-10-04
React Week 2

scott-ad-riley:
This is a rewrite of the second lesson to provide a recap of the basic concepts of React, with a view to then later (in the same lesson) having them start on some problems where they are just given requirements/a working demo and asked to implement it.

It walks through components, then props and then state. I've mentioned a functional component in the props part but all other snippets are using the `class` syntax. I'm looking for the functional/smart vs. dumb components to be a later refactoring - it's a common pattern, not a part of the framework.

I will update this PR as I gather those requirements/demos together - if you have any ideas just post them here or in slack and I'll include them. I'd like to have at least 4 or 5, up to about 10-15 with the intention of letting students pick and choose the ones they'd like to work on (so they should be at most 20-30 mins for a normal react developer - up to 1h/1h30 a student).

I've totally rewritten `lesson2.md` file - [go here](https://github.com/scott-ad-riley/syllabus/blob/week_2_notes/react/lesson2.md) to read it standalone instead of the diff view.

### DIFF
--- a/react/lesson2.md
+++ b/react/lesson2.md
@@ -1,82 +1,170 @@
-# React 2
+# React Lesson 2
 
-**What we will learn today?**
-- More Props and State
-  - Stateful (smart) vs Stateless (dumb) components
-- Component life cycle
-  - componentDidMount: Making API calls
-- Testing React
+In this lesson we're going to go through a recap of props and state, then we'll hand out a set of small tasks/apps to build in the browser. Most of them should take anywhere from 30 mins to an hour, you won't be expected to finish them all but you will not be given much starting code.
 
-In this lesson we will be learning to use React components more effectively with deeper understanding of `Props` and `State`.
+## Recap
 
-# Component Life Cycle
+### Component
 
-![First Render](/images/firstrender.png "Life Cycle Methods on the first render of a component")
-
-Components have `life cycle` events which are methods that run code at particular times in the process. According to the time that the methods are called, some of them are often prefixed with `will` and `did`.
-
-The diagram below shows the full picture of what happens during the rendering of a component. Even though you have all these methods available to use at your disposal, the only `required` method is `render()`.
-
-![Component Lifecycle](/images/componentlifecycle.jpg "All lifecycle Methods of a React component")
-
-Mounting
---
-These are called when a component is being mounted on the DOM such as:
-
-- Constructor()
-- componentWillMount()
-- render()
-- etc
-
-Updating
---
-When there are changes to the props or state:
-
-- componentWillReceiveProps()
-- componentDidUpdate()
-- render()
-- etc
-
-Unmonting
---
-When a component is being removed from the DOM:
-
-- componentWillUnmount()
-
-Other APIs
---
-These include `setState()` and `forceUpdate()`. You cannot access the state in some lifecycle methods, refer to the diagram above to find out which.
-
-The complete list of these methods with their descriptions can be found in the documentation [here](https://facebook.github.io/react/docs/react-component.html).
-
-
-## API calls
-
-In order to make an API called you can call the functions same
+At the highest level in React - you have components. These are created usually with 1 per file using the `class` syntax, like this:
 
 ```jsx
-callAPI = () => {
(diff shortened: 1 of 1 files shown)


extwiii:
I think examples are really good, especially adding/removing counter dynamically. This will give them some idea how the state works. Also, we should mention about immutable like this article ;
https://medium.com/@housecor/handling-state-in-react-four-immutable-approaches-to-consider-d1f5c00249d5
I just found a diagram explain react life cycle, please see the latest post;
https://stackoverflow.com/questions/30668326/what-is-the-difference-between-using-constructor-vs-getinitialstate-in-react-r
Also, their official examples also really good, we might go through with students to give some idea what React.js can do.
https://reactjs.org/

alismith-dev:
Looks good 👍

react/lesson2.md
```diff
@@ -1,82 +1,168 @@
...
-Once the state is updated with the data, your component will re-render.
+That's almost it for the recap, before we jump into the next part - have a go at trying to extend this counter example. Here's a few ideas that you could try:
 
-# Homework
-* [Lesson 2 Homework](Homework/Homework2.md)
+1. Change the colour of the counter, based on whether or not it's an even number ([demo here](https://csb-jopk2757y-mkxkadmfba.now.sh/))
+2. Add another button that multiplies the number by 2 ([demo here](https://csb-24pk29x2pn-zpnhuedswr.now.sh/))
+3. Add a slider that modifies that multiplier ([demo here](https://csb-l4kk6zn4kz-ypkmbtwnxy.now.sh/))
+  * So moving the slider to the left makes the multiplier 2, but moving it further to the right will increase the multiplier to 2,3,4 etc.
+4. Make the number of counters dynamic ([demo here](https://csb-nr8nvm9k70-fdvdaqcbst.now.sh/))
```
rarmatei: Are those URLs permanent?

rarmatei:
Really like the recap + exercises.

rarmatei:
Sorry, closed by mistake, re-opened :))

Really like the recap + mini-exercises idea. Maybe we can add introduce the lifecycle methods towards the middle of the day, and have them use them in the projects they built.

What are we planning for homework?

react/lesson2.md
scott-ad-riley: Yeh - in theory anyway. They could go after ~6 months if they don't receive any traffic, it depends on how http://zeit.co/now handles that sort of thing.

scott-ad-riley:
I'm thinking that we introduce ajax as a way to use lifecycle methods - possibly towards the end of the lesson. I'm going to create some notes tonight and add them in to this PR.

I'm hesitant to introduce them until the students are comfortable with the way components are updated/rerendered etc.

react/lesson2.md
```diff
@@ -1,82 +1,168 @@
...
 
+```jsx
+import React, { Component } from 'react'
+import Profile from './Profile'
+
+class MyComponent extends Component {
+  render() {
+    return (
+      <div>
+        <Profile name="Scott" age=24 />
```
djgrant: `{24}` or "24"

react/lesson2.md
```diff
@@ -1,82 +1,168 @@
...
 
-In order to use this function when the component first loads we will call it in the `componentDidMount()` lifecycle method.
+Notice how to get access to those props we ask for them off of `this`, and then we wrap them inside curly braces `{}` to tell jsx that we want it to actually evaluate those javascript variables.
+
+The above component could have also been written as a functional component, both are valid but we'll be sticking to the `class` syntax for the rest of the examples. Here is the same component defined using the function syntax:
 
 ```jsx
-componentDidMount() {
-    this.callAPI();
+import React from 'react' // don't need to import Component
```
djgrant: Might be worth adding a note that react is imported because JSX requires it.

react/lesson2.md
djgrant: I'm getting a build error on https://csb-24pk29x2pn-zpnhuedswr.now.sh

djgrant (approved):
Spot on, Scott 💯

PR #168, opened 2017-11-26
Fix code examples

alismith-dev:
Fix a couple of wrong/confusing code examples

### DIFF
--- a/js-core-2/week-08/lesson.md
+++ b/js-core-2/week-08/lesson.md
@@ -50,8 +50,8 @@ as an argument to query for elements matching all classes.
 // get an Element with id 'submit'
 var form = document.getElementById('myform');
 // get an HTMLCollection containing elements with class 'red'
-// which are children of the 'myform' element
-var inputs = root.getElementsByClassName('red'); 
+// which are children of the 'form' element
+var inputs = form.getElementsByClassName('red'); 
 ```
 
 `getElementsByClassName` can be called on individual `elements` as well as the
@@ -60,8 +60,8 @@ top-level `document` object. When calling `getElementsByClassName` on an
 the entire `document`
 
 ```js
-elements = document.getElementsByTagName('div'); // or:
-elements = someElement.getElementsByTagName('a');
+var elements = document.getElementsByTagName('div'); // or:
+var elements = someElement.getElementsByTagName('a');
 ```
 
 Much like `getElementsByClassName`, `getElementsByTagName` allows us to query


kkarimi (approved):

docs/js-core-1/week-1/lesson.md
```diff
@@ -25,6 +23,8 @@ By the end of this class, you should be able to:
 - List and describe the different primitive data types in JS
 - Use `typeof` to find the type of a variable
 - Assign data to variables using `let` and `const`
+- What are expressions and statements (and what's the difference)?
+- Define what expressions and statements are and describe the difference
```
illicitonion:
```suggestion
- Define what expressions and statements are and describe the difference
```

These seem the same

docs/js-core-1/week-1/lesson.md
```diff
@@ -25,6 +23,8 @@ By the end of this class, you should be able to:
...
 - Write, call and evaluate functions in JavaScript
```
illicitonion:
```suggestion
- Write and call functions in JavaScript
```

I'm not sure what evaluate is referring to here... Maybe the playing-computer style stuff? But I don't think we do that in this lesson?

docs/js-core-1/week-1/lesson.md
```diff
@@ -189,23 +188,16 @@ const product = 10 * 2; // 20
...
-2. Log a message that displays the total number of students and mentors
-
 #### Expected result
 
 ```sh
 Number of students: 15
 Number of mentors: 8
 Total number of students and mentors: 23
 ```
+### Numbers as decimals
```
illicitonion: This new block means there's quite a bit of space between the un-explained "Expected result" above, and the problem statement below - can we move the "Expected result" section down?

docs/js-core-1/week-1/lesson.md
```diff
@@ -239,6 +233,106 @@ Using online documentation, what other things can you do with the `Math` library
 
 Pick one thing on your table that you can do other than `Math.round` and prepare an explanation for the rest of the class.
 
+
+You can also use expressions with string interpolation or as a `return` value.
```
illicitonion: We haven't introduced functions/return/expressions yet, so maybe drop this section, or move it to later in the lesson?

docs/js-core-1/week-1/lesson.md
```diff
@@ -239,6 +233,106 @@ Using online documentation, what other things can you do with the `Math` library
...
+A statement is some code that typically carries out an instruction. You have encountered statements already:
+
+```js
+const greeting = "Hello Code Your Future!";
+```
+
+The above piece of code is a **variable declaration**, which is also a statement. 
+This statement is just an instruction to store the value `"Hello Code Your Future!"` in the variable `greeting`.
+
+Statements end in a semi-colon, though they can be omitted in JavaScript.
```
illicitonion: Being pedantic, block statements (e.g. if statements) are delimited with braces rather than semi-colons... I'm not sure whether to soften this sentence to something like "Most statements end in a semi-colon", or just skip over this for simplicity...

That said... I'm not actually sure introducing the term "statement" is useful for the lesson? Like, we don't rely on it anywhere, and it's not practically useful knowledge... Maybe we should just remove the "Statement" section completely, and only explain expressions?

docs/js-core-1/week-1/lesson.md
```diff
@@ -239,6 +233,106 @@ Using online documentation, what other things can you do with the `Math` library
...
+let product = 2 * 4; // assign the expression 2 * 4 to a variable product
+let worldGreeting = "Hello" + "World"; // assign the expression "Hello" + "World" to a variable worldGreeting
+```
+
+
+### Exercise ( 15 mins )
+
+You can run `node` by itself, which will open a _node console_, also called a [Read–Eval–Print Loop (REPL)](https://www.tutorialspoint.com/nodejs/nodejs_repl_terminal.htm).
+
+This console allows you to enter code in the console line by line and is a great way of testing bits of code before writing it in a script.
```
illicitonion:
```suggestion
This console allows you to enter code in the console line by line and is a great way of testing bits of code before writing it in a script. Each time you press enter, it will print out what the expression you typed evaluates to.
```

docs/js-core-1/week-1/lesson.md
```diff
@@ -239,6 +233,106 @@ Using online documentation, what other things can you do with the `Math` library
...
+if (true) {}
+
+const hoursInADay = 24
+
+hoursInADay * 7
+
+`My favourite colour is ${favouriteColour}`
+```
+
+(To exit the node REPL, you have to click Ctrl+d or Cmd+D on Mac)
```
illicitonion:
```suggestion
(To exit the node REPL, you have to click Ctrl+d)
```

It's actually still Ctrl+d on Mac - Cmd+D will duplicate your terminal into a split pane (or so I just found out by trying!)

docs/js-core-1/week-1/lesson.md
```diff
@@ -251,80 +345,143 @@ function double(number) {
...
 console.log(result); // 4
 ```
 
-The input given to a function is called a **parameter**.
+The function `double` is being invoked by placing parathenses after the name of the function like this `double()`
+
+🔑
+The input given to a function is called a **argument**
+
+🔑 A **parameter** is a variable that is used to refer to an **argument**
```
illicitonion, line 353:
```suggestion
🔑 A **parameter** is a variable that is used to refer to an **argument**

In the example above, we're saying that inside the function `double`, there is a variable called `number`, but we don't know what the value of the variable is - when someone calls the function, they will give a value for that variable.
```

docs/js-core-1/week-1/lesson.md
```diff
@@ -25,6 +23,8 @@ By the end of this class, you should be able to:
 - List and describe the different primitive data types in JS
```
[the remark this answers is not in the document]
alismith-dev: > We don't actually teach this at all

To me this is the more salient point: we shouldn't really have a LO that we don't actually fulfil.

docs/js-core-1/week-1/lesson.md
alismith-dev: I would also lean towards the latter. I try to follow [Teaching Tech Together's advice](https://teachtogether.tech/en/index.html#s:process-objectives) that LOs should be (somewhat) objectively answered: can the trainee demonstrate this knowledge or not.

docs/js-core-1/week-1/lesson.md
```diff
@@ -92,13 +92,13 @@ We can use a _variable_ to create a reference to a value. Variables can be thoug
 Before you use a variable in a JavaScript program, you must declare it. Variables are declared with _let_ and _const_ keywords as follows.
 
 ```js
-let greeting = "Hello world";
+let greeting = 'Hello world';
```
alismith-dev: Apologies if this is nitpicking, but I think we try to stick to the default Prettier settings since that is what trainees would have set up on their systems.

The default settings are for double quotes, not single (see in the [playground](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAbOMAEBzATnDASym0wF5MByACTlVQkwHcJdUATSgbhABoQIABxiFoAZ2SgAhrlwQmABRkIJKKaiZSAnhP4AjXFLABrDAGVBR4tmQxcAVzj84AWz1x27DwBkpJe1LYcABirC5SMCIkyCBS9jAQfCAAFjAuqADqyYTwYpZgcGYqOYQAbjlaMWBiuiDEYnC4MAqG2OHIAGbqDfwAVmIAHgBChibmUi5w3sRwnd1OIP0DZtboAIr2EPBzqD0glrgNuDF6Uu6oSYK4xDAZhOwwycgAHAAM-FcQDRmGgjFXcCOpVm-AAjpt4C0hKpYmIALRQAhedhJfDgwj4FqBdpILq7BYNFyEWwOAmrOAbLazXHzfgwM53B5PJAAJjphkIqGsAGEIC4cSBAQBWJL2BoAFTOqjxe1KjgAklAvLAzGBrsIAIJKswwLToHYNAC+hqAA)).

Perhaps you have some personal Prettier settings that are overwriting?

docs/js-core-1/week-1/lesson.md
```diff
@@ -189,23 +188,16 @@ const product = 10 * 2; // 20
...
 Number of students: 15
 Number of mentors: 8
 Total number of students and mentors: 23
 ```
+### Numbers as decimals
 
-## Numbers as decimals
-
-Numbers can be integers (whole numbers) or floats (numbers with a decimal).
+Numbers can be integers (whole numbers) or floats (numbers with a decimal point).
```
alismith-dev, line 193: I wonder if we should change the section header to "Numbers as floats" (or "Floating point numbers")? Or is that getting too much into details?

docs/js-core-1/week-1/lesson.md
```diff
@@ -239,6 +233,106 @@ Using online documentation, what other things can you do with the `Math` library
```
alismith-dev, line 235: Can't comment on it since it hasn't changed, but there's a typo on line 230 above: `Intructor` should be `Instructor`

docs/js-core-1/week-1/lesson.md
```diff
@@ -239,6 +233,106 @@ Using online documentation, what other things can you do with the `Math` library
 
 Pick one thing on your table that you can do other than `Math.round` and prepare an explanation for the rest of the class.
```
alismith-dev: Apologies, I feel like I'm nitpicking things that are already in the lesson.

This line is confusing: I think it is referencing a table from in-person lessons, and getting students on a table to work together. Would something like this work?

```suggestion
Working in small groups, pick one other thing that `Math` can do and prepare a short (2 - 3 min) explanation for the rest of the class. 
```

docs/js-core-1/week-1/lesson.md
```diff
@@ -239,6 +233,106 @@ Using online documentation, what other things can you do with the `Math` library
...
+function double(num) {
+  return num * 2; // expression num * 2 being returned from this function
+}
+```
+
+## Statements and expressions
+
+In JavaScript, there are **expressions** and **statements**. 
+We will use these words frequently to describe code.
+### Statement
```
alismith-dev: What do you think of rephrasing a bit to give more context?

```suggestion
In JavaScript, sections of code can called **expressions** and **statements**. We will use these words frequently to describe code.

### Statement
```

(I also cleaned up some of the newlines - unsure what your intention was though.)

docs/js-core-1/week-1/lesson.md
```diff
@@ -239,6 +233,106 @@ Using online documentation, what other things can you do with the `Math` library
...
+
+A statement is some code that typically carries out an instruction. You have encountered statements already:
+
+```js
+const greeting = "Hello Code Your Future!";
+```
+
+The above piece of code is a **variable declaration**, which is also a statement. 
+This statement is just an instruction to store the value `"Hello Code Your Future!"` in the variable `greeting`.
```
alismith-dev: Perhaps we should try to leave out words like "just" out? I think it reads cleanly without it?

```suggestion
This statement is an instruction to store the value `"Hello Code Your Future!"` in the variable `greeting`.
```

docs/js-core-1/week-1/lesson.md
```diff
@@ -10,22 +10,20 @@ import Feedback from "@theme/Feedback";
...
 
 - Use `console.log()` to print information to the console
-- List and describe the different primitive data types in JS
 - Use `typeof` to find the type of a variable
 - Assign data to variables using `let` and `const`
-- Write, call and evaluate functions in JavaScript
+- Define what expressions and statements are and describe the difference
+- Write and call functions in JavaScript
 - Manipulate strings with concatenation and interpolation techniques
 - Manipulate numbers with mathematical operators using the `Math` library
```
illicitonion:
```suggestion
- Manipulate numbers with mathematical operators and using the `Math` library
```

docs/js-core-1/week-1/lesson.md
```diff
@@ -203,41 +216,114 @@ Number of mentors: 8
...
-Numbers with decimals can be rounded to the nearest whole number using the `Math.round` function:
+We can take the value produced by an expression and assign it to a variable.
 
 ```js
-const preciseAge = 30.612437;
-const roughAge = Math.round(preciseAge); // 31
+let total = 1 + 1; // assign the expression 1 + 1 to a variable total
+let greeting = "hello"; // assign the expression "hello" to a variable greeting
+let product = 2 * 4; // assign the expression 2 * 4 to a variable product
+let worldGreeting = "Hello" + "World"; // assign the expression "Hello" + "World" to a variable worldGreeting
```
illicitonion: I think this might be clearer as:
```suggestion
let total = 1 + 1; // assign the value 2 (the value produced by the expression 1 + 1) to a variable total
let greeting = "hello"; // assign the value "hello" (the value produced by the expression expression "hello") to a variable greeting
let product = 2 * 4; // assign the value 8 (the value produced by the expression 2 * 4) to a variable product
let worldGreeting = "Hello" + "World"; // assign the value "HelloWorld" (the value produced by the expression "Hello" + "World" to a variable worldGreeting)
```
to emphasise that once an expression is evaluated, it doesn't matter how we got there, it has that value now

docs/js-core-1/week-1/lesson.md
```diff
@@ -251,80 +337,142 @@ function double(number) {
...
 ```js
 function add(a, b) {
   return a + b;
 }
 ```
 
-When you write a function (sometimes called _declaring a function_) you assign names to the parameters inside of the parentheses (`()`). Parameters can be called anything.
+In the function `add`, the parameters are `a` and `b`.
+
+If we call the function like this `add(10, 32)` then we're passing 2 arguments (or inputs) into the function : `10` and `32`.
```
illicitonion:
```suggestion
If we call the function like this `add(10, 32)` then we're passing 2 arguments (or inputs) into the function : `10` and `32`. We can use any expression as an argument.
```

docs/js-core-1/week-1/lesson.md
```diff
@@ -251,80 +337,142 @@ function double(number) {
...
+```
+
+The functions above behave similarly but _differ_ in some important ways.
+
+In groups, study the code above and then answer the following questions:
+
+a) How many times is the function `printMessage` called ?
+b) How many times is the function `getMessage` called ?
+c) What is the parameter name for the function `printMessage` ?
+d) What is the parameter name for the function `getMessage` ?
```
illicitonion: These aren't rendering as bullet points, they're all wrapped on one line - you may need to use numbers instead of letters here?

docs/js-core-1/week-1/lesson.md
```diff
@@ -251,80 +337,142 @@ function double(number) {
...
+
+Remember you can use `console.log` to log variables in order to check your answers to f) and g)
+
+Hint: Think carefully about the parameters and arguments for `getMessage` for part g)
+
+### Exercise (10 minutes)
 
-1. Write a function that returns the year someone is born given their age as input
-2. Using the answer from step 1, write a function that takes someone's name and age as input and returns a string that states the person's name and year they were born in a sentence
+1. Swap your laptop with your neighbouring pair and review each other's code for the `formatPenceToPounds` and `increaseByHalf` exercise - what can be improved? Is the code readable?
```
illicitonion:
```suggestion
1. Swap your laptop with your neighbouring pair (or send your code to them over slack) and review each other's code for the `formatPenceToPounds` and `increaseByHalf` exercise - what can be improved? Is the code readable?
```

PR #172, opened 2017-12-02
Add new page about making a PR

alismith-dev:
I found that we had several issues with students not being able to create PRs. So I wanted to document it and some of the ways of working around common mistakes.

Possible that the last section about actually creating the PR is unnecessary and we should just link to [the Github docs](https://help.github.com/articles/creating-a-pull-request-from-a-fork/). What do you think?


kkarimi (approved):
awesome!

PR #174, opened 2017-12-08
Draft week 10 and week 11 lessons

Buroni:
I changed a lot regarding the good code practices section in week 10, and rewrote week 11 to fit in with what we'll be teaching with class 3. Please give feedback about any changes/improvements you think should be made!

The reason I changed the "Example of bad code" in week 10 is because the previous one wasn't
actually runnable on its own (it contained a mysterious `uploadToDb` function). I'd like to show an
example of a bad standalone function which the students can run and try to improve themselves.


alismith-dev (approved):
Looks great! Thanks for your hard work on this!

PR #175, opened 2017-12-10
Clean up week 10

alismith-dev:
Move around resources & pull table of contents up to top for consistency with JS Core 2 weeks


illicitonion:
The main things we're lacking are:
* A survey of employers to see what kind of technical interviews they ask, and what their expectations are.
* Working out when we want to teach this, and therefore what knowledge we can assume.

I could start putting together roughly a workshop with the following assumed answers to those questions, which we could then augment/expand when we get better answers:
* Assuming discussing, whiteboarding, or online-code-writing a pure JS problem (i.e. the problem scope doesn't involve UI, frameworks, talking to APIs, etc) without any pairing
* Assuming a decent understanding of all of the content of JS1-3 (which realistically means teaching the workshop at least after the React class, because that's where a lot of the understanding of those modules is solidified)

I'd also suggest that this work would ideally depend on us having solved https://github.com/CodeYourFuture/syllabus/discussions/244 as well as some of the testing work that's underway, but we can develop these things in parallel (or flesh out this workshop once those are done) if we need :)

PR #178, opened 2018-01-06
Week five homeworks

robota-x:
**Still a work in progress**
Updates to lesson 5 (Manchester). Better description of changes is TODO, plus possible expansion of the Objects section.

### DIFF
--- a/js-core/week-05/homework.md
+++ b/js-core/week-05/homework.md
@@ -1,44 +1,101 @@
 # Homework
 
-1. [The repo](https://github.com/Code-Your-Future/JS-Core-1-Exercises) that you
-   have forked during the last class contains few challenges - solve them all in
-   JavaScript. You should know all the needed information by now.
 
-2. Follow
-   [this course](https://www.khanacademy.org/computing/computer-programming/programming)
-   on Khan Academy. It will go through some of the basics that we covered in the
-   class and beyond.
+### Challenges
+Go back to [the repo](https://github.com/Code-Your-Future/JS-Core-1-Exercises)
+that you have forked during the last class:
+* If you haven't attempted all the week 1 base challenges, go back and work on
+them!
+* If you're all done, challenge yourself with the week 1 advanced challenges,
+and the ones present in week 2.
 
-3. **Research:**
 
-Read the chapter from JS: The Good parts
+### Code style and best practices
+As you might have noticed, there are often several _valid_ ways to write the
+same code.
 
-* Chapter 9: Style - it's a short chapter, everyone start by reading that
-  chapter.
-* ALL: Read the chapter about **Objects** (up to Reference)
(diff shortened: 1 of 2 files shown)


js-core/week-05/homework.md
```diff
@@ -1,44 +1,101 @@
...
+Consistency and predictability in the _visual structure_ of the code won't
+change the way the computer will execute the code, but will make it way easier
+to maintain the code and spot bugs for both you and other programmers.
+
+To this end, many organisation and people develop a set of _conventions_ ,
+or **style guide** on how to structure their code. While often one option is not strictly 'better' than others, it is tremendously helpful if all the people
+working on a project write in the same style.
+
+To get some more insight on the argument, read these links:
+* https://www.smashingmagazine.com/2012/10/why-coding-style-matters/ dig a bit
```
kkarimi, line 60: +1 for content though it's nicer to not drop the full link in the text but use its title and link

js-core/week-05/homework.md
```diff
@@ -1,44 +1,101 @@
...
+complicated, and it's increasingly difficult for developers, especially juniors,
+to have a mental map of every single line of code. Every change on the codebase
+risks breaking the functionality in unexpected ways leading to a never-ending
+loop of bugs.
+
+One of the ways to mitigate this is testing: in essence, writing extra code,
+often with the aid of specialised tools, to validate the functionality of the
+actual program.
+
+This will be a subject in future lessons and there are several other good
```
kkarimi, line 94: +1

js-core/week-05/lesson.md
```diff
@@ -12,66 +12,97 @@
...
-things, identifiers like `name1`, `name2`, `name3` and so on quickly become
-tiresome. **Arrays** to the rescue!
+If you want to store a **list of values**, declaring manually a number of
+varibles becomes quickly tiresome and untenable:
 
-Imagine an array like a list or a collection of variables, but instead of each
-variable having its own name, only the array has a name. So how do you access a
-variable inside, then? How do you read and write? Using an _index_!
+```js
+var name1 = 'Bill'
```
kkarimi, line 19: Every time we do/update this class we should consider, is it time to switch this to `const`/`let` yet? answer is probably still no.

robota-x: I avoided es6 as I saw it being introduced as a lesson in week 11 so didn't want to generate more duplication.

More generally, many languages introduce the concepts of const and var at a early stage, so I don't it should be too hard to start from those directly. I the relation between `let` and `var` is somewhat more subtle and needs a decent understanding of variable scope to avoid confusion..maybe introduce `const` and `var`, and have `let` only appear after scope? There is the downside of most es6 (that I've seen) using `let` as the default var keyword tho.

If anything, a good argument to keep var might be that the vast majority of stack overflow questions and answers on javascript uses vanilla JS.

Issue #183, opened 2018-01-15
JavaScript III: Week 10 - Good Code, TDD (additional material)


christopherallanperry:
As a way of giving the students additional practice for TDD, I'd like to propose that we include Marc Littlemore's (BBC) free 7-lesson [http://www.marclittlemore.com/courses/javascript-testing-made-easy/]("JavaScript Testing Made Easy") course in the homework section of Week 10.

It makes use of Mocha/Chai/Sinon rather than Jest, but it will give the students some exposure to alternative testing frameworks, show them that these have a great deal of commonality, and should be nothing to be concerned about.

I have spoken to Marc on the use of the course for CYF and he was more than happy that we make use of it. It includes a daily email for each lesson with a clear set of instructions to follow, a set of YouTube videos and an associated GitHub repository.

christopherallanperry:
Thanks for merging this @kkarimi - I'm not seeing it appear on the live syllabus though, but can't see an obvious reason why that would be. Can you have a look?

kkarimi:
@christopherallanperry no prob, looks like one of the commits broke the build yesterday which meant it did not publish. It's [there](https://codeyourfuture.github.io/syllabus-master/js-core-3/week-10/lesson.html#homework) now

christopherallanperry:
@kkarimi : Thanks Kash

PR #189, opened 2018-01-18
Modified TDD instruction for Week 10 homework - Issue #183

christopherallanperry:
Additional item added to the Lesson 10 homework. Students to register and work through the Marc Littlemore "JavaScript Testing for Beginners" course.

### DIFF
--- a/js-core-3/week-10/homework.md
+++ b/js-core-3/week-10/homework.md
@@ -1,15 +1,16 @@
 # Homework
 
 1. Find examples in your own code where you violated the three principles of
-   good code (Clarity, Reusability, Extensibility). For each principle, find one
-   example.
+good code (Clarity, Reusability, Extensibility). For each principle, find one
+example.
 1. Clone the repo of the
-   [election project](https://github.com/codeyourfuture/cyf-election) and finish
-   all of the part 1 exercises.
+[election project](https://github.com/codeyourfuture/cyf-election) and finish
+all of the part 1 exercises.
+1. Register yourself on the Marc Littlemore's free [JavaScript Testing for Beginners](https://www.marclittlemore.com/courses/javascript-testing-beginners-course/?signup=cyf-course) course and work your way through the 7 lessons, which are delivered via a daily email. The course uses a Mocha/Chai framework, which is in common use within the web development industry. This will give you more TDD practice using a different approach to Jest, as well as more exposure to ES2015/ES6 JavaScript.
 
 ## Research
 
 1. Research other module formats than CommonJS. What is AMD? What are ES6
-   modules and how do their differ from CommonJS?
+modules and how do their differ from CommonJS?
 1. What are other test frameworks for JavaScript?
 1. More parts of the Jest (Jasmine) DSL than just `.toBe()`


kkarimi (approved):

docs/js-core-2/cowsay-project/readme.md
```diff
@@ -0,0 +1,120 @@
...
+Cowsay was originally written in [Perl](https://simple.wikipedia.org/wiki/Perl). In this coursework, we will run it with Node. How would we go about this? What things would be helpful? I think we could use:
+
+* A node package that contains the cowsay library. I'll start you off by letting you know there is one, and it is called ```cowsay```
+* A way to execute the script.
+
+Before we start, what's the fastest way you could get the cow to talk? Well, you can execute some packages directly [in the command line with npx](https://nodejs.dev/learn/the-npx-nodejs-package-runner).
+
+```
+npm install cowsay
+npx cowsay Moooo
```
alismith-dev: A handy trick with `npx` is that it will (temporarily) install the package that is requested before running it. So this line can be:

```suggestion
npx cowsay Moooo
```

docs/js-core-2/cowsay-project/readme.md
```diff
@@ -0,0 +1,120 @@
...
+```
+
+We did it! But let's explore some more. How would you install cowsay as a dependency? Make a new folder called /cowsaying and initialise a new package.
+```
+mkdir /cowsaying
```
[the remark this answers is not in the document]
alismith-dev: If we added some instructions to navigate to the trainee's CYF work folder, would `mkdir cowsaying` work? I feel like adding to `~` would not be helping with the folder mess.

docs/js-core-2/cowsay-project/readme.md
```diff
@@ -0,0 +1,120 @@
...
+cd /cowsaying
+npm init
+```
+Press return through all the options and then do.
```
alismith-dev, line 42: Hmm, I was about to suggest that we remove this, to avoid the interactive prompt. However I had forgotten that `npm install X` without a prior `npm init` will a) output a bunch of scary looking warnings and b) not create a `package.json`.

So instead, maybe we could switch this to `npm init -y`? This runs the same init command but answers yes to all the questions (which is fine since we don't expect to publish this).

docs/js-core-2/cowsay-project/readme.md
```diff
@@ -0,0 +1,120 @@
...
+```
+npm install cowsay
+```
+### npm run
+
+Let's find out how to execute cowsay from our package.json. Open /cowsaying in your text editor and look at the [package.json](https://nodejs.dev/learn/the-package-json-guide).
+
+Add cowsay to the scripts array.
```
alismith-dev: Nitpick: it's an object not an array 🙂 

```suggestion
Add cowsay to the scripts object.
```

docs/js-core-2/cowsay-project/readme.md
```diff
@@ -0,0 +1,120 @@
...
+
+Remove the hardcoded "Mooooo".  Can you pass a different string in the command line?
+
+### node 
+
+Now let's execute cowsay in a different way. Open /index.js in your text editor. Import the cowsay package with require.
+```
+const cowsay = require("cowsay");
+```
+And log a cow to the console that says "I require a package". Add some more arguments if you can.
```
alismith-dev, line 70: Perhaps it might be useful to link to the [cowsay documentation](https://www.npmjs.com/package/cowsay) here?

alismith-dev: Actually... having poked around with it a bit, I think the docs are pretty confusing (and maybe plain wrong?). So I'm having second thoughts about it.

docs/js-core-2/cowsay-project/readme.md
```diff
@@ -0,0 +1,120 @@
...
+```
+console.log(cowsay.say({
+        text : "I require a package",
+        e : "oO",
+        T : "U "
```
alismith-dev, line 75: I found these options a bit confusing when I first saw them.

The docs seem to imply that it accepts `eyes` and `tongue` as options, but having looked in the source I don't think that's actually the case :/ Perhaps we could introduce them as a separate step with an explanation of what they mean?

docs/js-core-2/cowsay-project/readme.md
```diff
@@ -0,0 +1,120 @@
...
+What are the downsides of this approach? I would say one is that we want to be able to pass in arguments, not just always say moo. How could we make the cow say Baaaa?
+
+Remove the hardcoded "Mooooo".  Can you pass a different string in the command line?
+
+### node 
+
+Now let's execute cowsay in a different way. Open /index.js in your text editor. Import the cowsay package with require.
+```
+const cowsay = require("cowsay");
+```
```
alismith-dev, line 69: Minor nitpick, you get nicer syntax highlighting if you mark the code block as Javascript by putting `js` just after the initial three backticks (I'd give an example but it's painful to add markdown examples in markdown)

PR #204, opened 2018-02-04
[WIP] React

alismith-dev:
WIP rewrite/port of React module

Things I'm not sure about:

- I'd quite like to move the "Reacting to Changes" section into the first week but was worried we'd run out of time. For the last London cohort I don't think we did a good job of communicating what React is **for**. For example - how is it different from Handlebars? Imo, showing simple dom updates (e.g. the counter example) is a good way to demonstrate this
- If we did move the "Reacting to Changes" section into the first week, we could remove references to props and just hold the counter as global state. This would avoid having to cover props in the first week
- Most of the example code is not very relevant. I couldn't think of better ideas to implement. Suggestions welcome 🙂 
- Suggestions on further reading/watching material also welcomed

Some reasoning behind some of the decisions I made:

- Wanted to make some of the "magic" of CRA more explicit so started off with just loading the react deps from unpkg.org in a `<script>` tag. Then move onto CRA later
- Some of the students in the last London cohort really struggled switching between functional and class components. Since state requires a class component, I think it's better to use that as the default. We can add in functional components later, or they can pick it up themselves

TODO

- Props & state
- Lifecycle & `fetch`
- Kash suggested we update the [cyf-hotel-react homework](https://github.com/CodeYourFuture/cyf-hotel-react) to be structured more like the [cyf-hotel homework](https://github.com/CodeYourFuture/cyf-hotel). Maybe we can just rewrite the exercises as tickets/user stories?


alismith-dev:
There's been some requests to merge this so it's easier to review - more than happy to make changes once it's on the syllabus page

PR #217, opened 2018-02-24
add example of using components in seperate files

dahfool:


dahfool:
@40thieves @kkarimi

alismith-dev:
LGTM thanks :D

PR #237, opened 2018-05-08
two fields needs to be added which is fixed

mahsa2017:


alismith-dev:
@crdoconnor Is this worth looking at while you're working on the db module?

bonboh:
The db content has been updated so these lines do not exist anymore

PR #239, opened 2018-05-08
conceptual typo fixed

mahsa2017:


crdoconnor (approved):

crdoconnor (changes requested):
This wasn't a typo. The whole point of a foreign key is that it will raise errors if invalid data is put in, which we want.

alismith-dev:
Closing due to inactivity.

PR #240, opened 2018-05-12
Update lesson.md

PaulPhillipz:
added a few small changes as a test. Showed link in full rather than just the word 'here', so that students can see where it will take them to. 

Thought about setting up a branch for "manchester" but did not for such a small change.


alismith-dev:
Great thanks 😄

grimxreaper:
Hi @ChrisOwen101 I'd like a stab at this as my first contribution if that's okay?

PR #251, opened 2018-08-20
PoC for putting exercises into syllabus

alismith-dev:
Proof of concept showing how we could move the js-exercises into the syllabus itself. This could solve the problems discussed [here]](https://github.com/CodeYourFuture/js-exercises/pull/134#issuecomment-403773339). Intended as a strawman


djgrant:
@40thieves Thanks for spiking this. I really like this option. It's a simple solution, easy to maintain (as opposed to two repos), brings all the content into one repo, and let's the student stay inside their IDE if they want. (It also keeps open the possibility of one day doing something to run the exercises in the gitbook).

The main issue as you've mentioned before is cluttering the syllabus repo with student's PRs. But we've also identified the process of creating PRs as being a hindrance at this stage so I'm wondering if we can fix that too. I'm not sure there's a quick solution. Maybe for now we could ask students to fork the repo and open PRs to their own master branch?

PR #254, opened 2018-10-17
Week-1 HTML: fix links

robota-x:
2 images were broken in the syllabus due to the path being wrong. I took the opportunity to also fix a couple of related broken links


alismith-dev (approved):
Looks good thanks

alismith-dev:
I think the tests are failing just because keys aren't set up correctly. Don't see anything here that would have broken it. Merging

PR #255, opened 2018-11-25
Improved state discussion

mcsdevv:
Added another example to show the difference between updating state when relying on an existing state value or not. I have added this as I came across a few examples where students were having issues updating state when calculating values based off existing state values.

### DIFF
--- a/react/week-20/lesson.md
+++ b/react/week-20/lesson.md
@@ -274,10 +274,9 @@ class Counter extends Component {
   }
 
   increment = () => {
-    let currentCount = this.state.count
-    this.setState({
-      count: ++currentCount
-    })
+    this.setState(prevState => ({
+      count: prevState.count + 1
+    }));
   }
 
   render() {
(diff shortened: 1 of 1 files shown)


alismith-dev:
Hey. Sorry that's its taken me a while to get around to looking at this.

Originally my thinking was that we should focus on just teaching the set state/re-render cycle and deliberately left out the updater function mechanics. For this explanation, I was trying to really focus on what happens in `render` when state is changed.

However on reflection, I think it is important for the students to know and I think your explanation is good. How would you feel about moving the updater function into a separate section? I realise that in the first example it would be technically incorrect, but we would then correct it in a subsequent section. I'm not 100% sold on this idea though, so if you disagree I'm happy to merge this.

mcsdevv:
I can see both sides, but I do feel it would be best if we could just avoid the 'technically' incorrect part completely.

I was going to suggest using another example because this is a common issue when using the counter approach but I can't think of a super simple way to illustrate it. I'll have a think over the next few days and come back to you on this.

alismith-dev:
> I'll have a think over the next few days and come back to you on this.

Thanks! I'll have a think too (although I'm super busy at the moment, so it make take a while 🙂)

mcsdevv:
I've had the chance to weigh up both approaches, both having it as part of the state discussion and having it as a separate section (passing in prevState).

I'm still happy with my initial approach here as I feel it fits well within the current section and is clear to follow and easy to understand. Maybe in the future (when time is available) a slightly different example could be utilised which would make it easier to avoid this kind of problem. Therefore, if you could merge that would be great, just let me know if you want anything else changed :slightly_smiling_face: :

alismith-dev:
I'm actually on the way into our teaching session at the moment so can't
link it properly but I've started a branch called #react-improvements which
implements your suggestion, with a little tweaking. Do you mind having a
look at that and seeing what you think?

mcsdevv:
Nice work, I've added a little more to the discussion so that students understand why we do this, not just when. Your explanations are more concise and read well, closing this PR after raising it on the new branch.

PR #256, opened 2018-12-07
Spike out mentors notes

alismith-dev:
As discussed on a call recently, we want to provide mentor's notes for each lesson. This PR does **not** contain any such notes, but spikes out an idea for being able to include mentor notes in the compiled html output. The key points are:

- The mentor guide is included as a top level link in `SUMMARY.md`. This is necessary to get gitbook to compile the markdown file into html and create a link to the page
- Because of the above, the mentors notes will appear in the navigation sidebar. To reduce confusion I've put the mentor guide link at the bottom
- Hopefully it's obvious that the mentors notes I've added is just a stub and will need to be extended with actual content 🙂 

I believe this hits the main constraints of:

- Being easily linkable
- Not being in another repo
- Easily editable (We believe that Github's inline editing tools may be good enough. Although this is yet to be tested on someone non-technical)

### DIFF
--- a/SUMMARY.md
+++ b/SUMMARY.md
@@ -54,4 +54,7 @@
 
 * [Contributing](CONTRIBUTING.md)
 * [Lesson Template](/others/lesson-template.md)
-* [HTML/CSS Mentors guide](/html-css/mentor-guide.md)
+* HTML/CSS
+  * [Mentors guide](/html-css/mentor-guide.md)
+* JavaScript I
+  * [Week 4 - Hello JavaScript](/js-core/week-04/mentors.md)
--- a/js-core/week-04/lesson.md
+++ b/js-core/week-04/lesson.md
@@ -2,6 +2,8 @@
 
 # JavaScript Core I - 1
 
+[Mentor Notes](./mentors.md)
+
 **What we will learn today?**
 
 * [Install Node](#install-node)
--- a/js-core/week-04/mentors.md
+++ b/js-core/week-04/mentors.md
@@ -1,0 +1,3 @@
+# Mentor's Notes
+
+Some notes about JS1-1


djgrant:
Nice one! Did you manage to get this rendering in gitbook? Any chance you could post a screenshot?

rarmatei:
Thanks for this!!

I know this is already getting very complicated, but I'm wondering if we could somehow create some links in-between: js-exercises/week1, syllabus/week4-lesson, syllabus/week4-mentor-notes.

We're moving towards having all 3 of the above as representing the same information but under different formats (js-exercises is very practical/class oriented, the syllabus is a compiled-together, easily readable set of notes from js-exercises, and the mentors notes is what it says it is).

That would also reduce the chance that they get out of date with each other, or we end up with copies because people didn't know about the existence of the other 2.

alismith-dev:
> Any chance you could post a screenshot?

I have gitbook locally installed, so yep:

Link from the teaching page:

screen shot 2018-12-08 at 19 00 40

Link in the sidebar:

screen shot 2018-12-08 at 19 00 47

alismith-dev:
> We're moving towards having all 3 of the above as representing the same information but under different formats

That's a very good point, I hadn't really considered.

I'm afraid I haven't really got a lot of good ideas at least in the short term. We may have to live with it for a bit. I don't feel like it's a blocker on getting a start on mentor notes.

In the longer term, there is this experiment https://github.com/CodeYourFuture/syllabus/pull/251, which tries to combine js-exercises & the syllabus repos together. There are several downsides to this however - the main one being that student's homework would potentially clutter up the syllabus repo. Feel free to suggest alternatives though. I don't think there's a clear winner in terms of solutions.

djgrant:
Looks great to me. I don't think we need extra links for class teachers.

martynamaron (approved):
Could we use this for class plan and ideas then, please? :)

jonny-no1 (approved):
Looks good to me. Us 4 mentors for london js1 will start populating it. Thanks @40thieves!

PR #258, opened 2018-12-22
Fix broken links and clean up duplicate files

jonny-no1:
⚠️ Don't merge yet  ⚠️
I've noticed that there are duplicate files that live both in the root folder and under `/other`. The menu of gitbook link to `/other` while some in-text links still point to the root files. I'll compare the files one by one and make sure that the latest version is in `/other` so that the duplicate files in the root folder can be removed.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -70,7 +70,7 @@ bullets under the main README.
 
 ## A Lesson template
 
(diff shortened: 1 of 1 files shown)


bonboh:
The duplicate files in the root and other directories have now been sorted

PR #259, opened 2019-01-06
Update lesson.md

mcsdevv:
Minor change, added the 'why' to setState callback. Previously it had an idea of when to do this, having a 'why' we do this should be beneficial for all.

### DIFF
--- a/react/week-20/lesson.md
+++ b/react/week-20/lesson.md
@@ -377,7 +377,7 @@ class Counter extends Component {
 
 Now we can count up as many times as we like! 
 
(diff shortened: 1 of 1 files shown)


alismith-dev:
Excellent point. Thanks!

alismith-dev:
And also apologies I meant to give you the heads up on my branch. Give a bit of background on my branch:

a)  I'm pulling in the feedback from the last time we did React (so if you have other feedback let me know, or add more PRs)
b) We're changing up the order of the syllabus here in London so that React comes before Node/DB and so a lot of the changes are to make the exercises lesson jargon-y and more step-by-step (previously they were deliberately written to be more like you would see in a job)

PR #262, opened 2019-01-09
Add Airtame installation instructions

lpalma:
As we had some Students facing issues while trying to install Airtame, I'm adding the installation instructions (for Ubuntu only) to Lesson 0.
We can move to another place if Lesson 0 is not the correct place to put it.


lpalma:
closing in favour of #265

PR #263, opened 2019-01-10
Fix explanations of sort() and join() array methods

martynamaron:
Once merged to `master`, it should also be `merged` into `london` branch.


martynamaron:
Closing - will raise the same change from a clone

PR #265, opened 2019-01-11
Add Airtame installation instructions

lpalma:
As we had some Students facing issues while trying to install Airtame, I'm adding the installation instructions (for Ubuntu only) to Lesson 0.
We can move to another place if Lesson 0 is not the correct place to put it.

### DIFF
--- a/lesson0.md
+++ b/lesson0.md
@@ -9,6 +9,7 @@
 * [Terminal](#terminal-basics)
 * [Git](#git)
 * [IDEs](#ide)
+* [Airtame](#airtame)
 * [Homework](#homework)
 
 ---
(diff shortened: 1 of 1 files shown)


alismith-dev (approved):
Looks good. Thanks for debugging and writing that up!

alismith-dev:
Ah, hmm I'm afraid this is actually in the wrong file (but the right lesson).

Unfortunately we seem to have 2 `lesson0` files: `./lesson0.md` and `./others/lesson0.md`. I believe Jonny was making an effort to clean this up in https://github.com/CodeYourFuture/syllabus/pull/258.

The `week0` file that is rendered out is actually `./others/lesson0.md`, because it is linked to from `SUMMARY.md` which is what gitbook uses to compile the project. If you could move these instructions to `./others/lesson0.md` instead then they'll show up. Sorry about the confusion!

PR #266, opened 2019-01-11
Change More Arrays in JS1 to week 3

martynamaron:


alismith-dev (approved):
Good spot! Thanks

docs/guides/intro-to-tests.md
```diff
@@ -0,0 +1,154 @@
...
+## What does this cover
+
+This introduction is designed to give you a high-level overview on the basics of testing, which you'll need when you're tackling the problems in your coursework at Code Your Future.
+In this introduction to testing we'll cover the following things:
+
+- Code specification
+- What is a test ?
+- Anatomy of a test ( using Jest )
+- Understanding test feedback ( using Jest )
+
```
alismith-dev, line 14: I unfortunately missed the previous syllabus team meeting, so this may have already been discussed. If that's the case, feel free to ignore me 🙂 

Having said that, I feel like we're missing an opportunity to discuss why testing is useful here. I don't think it needs to be exhaustive and I think we could make it fairly specific to why CYF trainees should learn it now.

Something along the lines of:

> **Why test?**
> Throughout the JS Core modules, we'll use tests in your coursework so it's important you know how to understand what a test is and how they work. The tests will:
>
> - Check whether you've solved the problem
> - Guide you towards the correct solution
>
> Tests are also commonly used by professional software developers, so it's useful to start learning early how to read and interpret them.

docs/guides/intro-to-tests.md
```diff
@@ -0,0 +1,154 @@
...
+In this case, we would expect the `output` to be:
+
+```js
+{
+    "north" : 2,
+    "by" : 1,
+    "west" : 1
+}
+```
+
```
alismith-dev, line 64: A couple of points here.

This feels like a good opportunity for an exercise:

> Write another example with a different `input` variable. What do you expect the `output` to be?

It also feels like a good time to explore why automation is useful.

Something along the lines of:

> We could keep trying out different versions of `input` and checking `output` but real software will often have thousands (or even millions!) of these cases. We can't keep checking all of these ourselves.
> This is where *automated* testing comes in: we could write some more code that checks whether the cases are working as we expect. Let's look at what an automated test looks like.

docs/guides/intro-to-tests.md
```diff
@@ -0,0 +1,154 @@
...
+![Failing test case for createWordCount](./assets/test-fails-case.png)
+
+Jest aims to give feedback that makes it simple to see why test has failed. It shows us several things:
+
+- the line number for the failing test (line 9)
+- shows us `expected` - which corresponds to our `expectedResult`
+- shows us `received` - which corresponds to our `actualResult`
+
+We can see the test feedback is telling us that `received` is `undefined`. In other words, `actualResult` ( `createWordCount`'s return value) is `undefined`.
```
alismith-dev: Should we say something about how to make the test pass here?

> To get this test to pass, we would have to make sure that `createWordCount` returns an object `{ hello: 1 }`

docs/guides/intro-to-tests.md
```diff
@@ -0,0 +1,154 @@
...
+
+```js
+const input = "hello";
+const actualResult = createWordCount(input);
+const expectedResult = {
+  hello: 1,
+};
+```
+
+In this section, we're declaring some important values that we will need to compare later on in the test.
```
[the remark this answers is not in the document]
Dedekind561: This was another thing I wrestled with - I purposely omitted the inline example but I do agree it is useful to see both so they understand that the assertion is the critical part. I also like the idea of starting with the inline example to emphasise the importance of the assertion and then moving on to another example with the variables. I want ( as much as possible ) for them to leave the guide with the knowledge that the code can take many forms as when I first learnt TDD I thought there always some fixed prescription for the test code.

PR #284, opened 2019-03-31
React tweaks

alismith-dev:
Some minor tweaks in response to feedback from @nbogie. Also added a section on controlled vs uncontrolled components (noted in the mentors notes that this can be dropped if short on time)

### DIFF
--- a/react/week-19/lesson.md
+++ b/react/week-19/lesson.md
@@ -357,7 +357,7 @@ const List = () => (
 )
 ```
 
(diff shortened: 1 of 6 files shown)


alismith-dev:
@nbogie If you want to have a look over this that would be really great :)

alismith-dev:
Going ahead with merging this @nbogie, hopefully there's nothing too objectionable in there

PR #286, opened 2019-04-28
Update first React class

llh1:
Few fixes based on our experience in Scotland CYF:

- We taught students to write JS with ';' at the end of each line, and create-react-app also uses this convention, so exercises and examples have been updated to use this as well.
- Use `querySelector` over `getElementById`.
- Typos and various small fixes

### DIFF
--- a/react/week-19/lesson.md
+++ b/react/week-19/lesson.md
@@ -163,13 +163,13 @@ import ReactDOM from 'react-dom';
 function HelloWorld() {
   return (
     <div>Hello World</div>
-  )
+  );
 }
 
 ReactDOM.render(<HelloWorld />, document.querySelector('#root'));
 ```
 
-There are 4 important parts in this code:
+There are 3 important parts in this code:
 
 1. First we import `React`. This is important because JSX is converted to `React.createElement` calls. If the `React` variable is undefined then this will fail.
 2. We create a React component called `HelloWorld`.
@@ -177,7 +177,7 @@ There are 4 important parts in this code:
 
 > **Exercise D**:
 > Using the `pokedex` React app that you just created and open the `src/App.js` file
-> 1. Delete everything in the file except the line containing `export default App`. You should see an error in your terminal and in your web browser - don't panic! We're going to remake the `App` component ourselves.
+> 1. Delete everything in the file except the line containing `export default App`. You should see an error in your terminal and in your web browser - don't panic! We're going to remake the `App` component ourselves
 > 2. Import React variable from the React package
 > 3. Create a function named `App`, which will be our component
 > 4. Within the `App` function, return a `<h1>` element with the text "Welcome to the Pokedex". What do you see in your web browser?
@@ -262,7 +262,7 @@ If we want to do this, we can still use arrow functions but we can't use the imp
 
 So far all of the components we have looked at haven't been able to change - they are *hard-coded*. But this doesn't make very interesting websites, we want to be able to use variables with different data. We can insert variables (and some other things) into our React components.
 
-Anything that is inside curly braces (`{` and `}`) is interpreted as a regular JavaScript *expression*. That means you can use every object or function from JavaScript that we have learned so far. Let's look at an example ([interactive example](https://codesandbox.io/s/o5vzzyxm9)):
+Anything that is inside curly braces (`{` and `}`) is interpreted as a regular JavaScript *expression*. That means you can use every object or function from JavaScript that we have learned so far. Let's look at an example ([interactive example](https://codesandbox.io/s/l910pqnjql)):
 
 ```js
 const Greeting = () => {
@@ -272,7 +272,7 @@ const Greeting = () => {
   );
 }
 ```
-Now instead of hard-coding the greeting in the `Greeting` component, we are using a variable. Remember that everything between the curly braces is just regular JavaScript. So we can use more than just variables ([interactive example](https://codesandbox.io/s/1016mnr7o7)):
+Now instead of hard-coding the greeting in the `Greeting` component, we are using a variable. Remember that everything between the curly braces is just regular JavaScript. So we can use more than just variables ([interactive example](https://codesandbox.io/s/nw29kzx744)):
 
 ```js
 const Mentor = () => {
@@ -315,7 +315,7 @@ const Name = () => {
 };
 ```
 
-A common pattern in React is to use `Array.map` to loop through a list of items and render a component for each one ([interactive example](https://codesandbox.io/s/8yp87yn7kl)):
+A common pattern in React is to use `Array.map` to loop through a list of items and render a component for each one ([interactive example](https://codesandbox.io/s/7mw0mw3qx0)):
 
 ```js
 const mentors = ['Ali', 'Kash', 'Davide', 'German', 'Gerald'];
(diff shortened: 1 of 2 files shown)


react/week-19/lesson.md
```diff
@@ -135,8 +134,7 @@ Using JSX ([interactive version](http://jsbin.com/gekahexige/edit?html,output)):
 As you can see, this is much easier to read than both the straight `React.createElement` API and the vanilla JS API. Most people using React use JSX to write their components.
 
 > **Exercise B**
-> 1. Change the [JSX example from above](http://jsbin.com/gekahexige/edit?html,output) to instead render a `h1` with an id of `"main-header"` and the text "Hello Ali"
-> 2. Visit the [Babel REPL](https://babeljs.io/repl/) and copy/paste the `h1` that you just created
```
alismith-dev, line 139: Perhaps it's just me, but I do quite like the ability for students to see that JSX isn't "magic", it's just regular functions. I feel like it reduces the "this is weird new syntax" aspect of it.

Having said that I felt like we spent too much time on it in London for the last class. So I think I'm ok with removing this exercise.

react/week-19/lesson.md
```diff
@@ -159,27 +157,28 @@ The last command should open a web browser for you. This shows the application t
...
 ```
 
-There are 3 important parts in this code:
+There are 4 important parts in this code:
 
-1. First we import `React`. This is important because JSX is converted to `React.createElement` calls. If the `React` variable is undefined then this will fail
-2. We create a React component called `HelloWorld`
-3. We *render* the `HelloWorld` component into a `div` with the id of `root`
+1. First we import `React`. This is important because JSX is converted to `React.createElement` calls. If the `React` variable is undefined then this will fail.
+2. We import `ReactDOM` which will be used to *render* the component on the screen.
```
alismith-dev: Interestingly `ReactDOM` kept coming up as a problem for class 5 in London.

In all of our interactive examples we have the `ReactDOM.render` right in the same file. Which is different to the student's CRA code - there's already a `ReactDOM.render` implemented in `index.js`.

They thought that it was important to have `ReactDOM.render` in every file. Additionally we use the word "render" a lot to mean returning a component, so it kind of makes sense that you would need a method named `render`.

I found it difficult to explain what `ReactDOM.render` was doing. And how it was different from just regular `React` without going off on a major tangent.

I'm leaning towards mostly ignoring it and changing all of the examples to have a split between the code that mounts onto the root node and the "application" code. Basically to mimic the default CRA file structure.

react/week-19/lesson.md
```diff
@@ -159,27 +157,28 @@ The last command should open a web browser for you. This shows the application t
 We looked at the beginning of the lesson at the concept of components. Now let's look at how components are made in React.
 
 ```js
-import React from 'react'
-import ReactDOM from 'react-dom'
+import React from 'react';
+import ReactDOM from 'react-dom';
 
 function HelloWorld() {
   return (
```
alismith-dev, line 164: Just a quick note on the parens here - I was deliberately trying to leave them in, because multi-line JSX can cause problems sometimes. 

I just tested this:

```js
return
  <div>
    <span>Hello world</span>
  </div>
```

Prettier will insert a semi after the return, which I think would trip up the students.

However this for this code:

```js
return <div>
  <span>Hello world</span>
</div>
```

Prettier does the right thing.  I feel like this would be the more common scenario for students to type, so in the end I don't think it really matters that much.

Overall I think it's something that's good to mention when showing a component for the first time.

react/week-19/lesson.md
```diff
@@ -189,19 +188,15 @@ There are 3 important parts in this code:
 
 #### Component Composition
 
-A component can be combined with another component so that both are rendered. This is sometimes called *composition* ([interactive example](https://codesandbox.io/s/0x4wonqn00)):
+A component can be combined with another component so that both are rendered. This is called *composition* ([interactive example](https://codesandbox.io/s/0x4wonqn00)):
```
alismith-dev, line 190: I can get these codesandboxes updated with those tweaks.

I suppose we could probably set up a group account or something so that they're not all under my account.

Also, while I'm on the subject - the codesandboxes are all configured to be "frozen", which means that if you edit them, and accidentally hit save then it will prompt you to fork. I find it **really** useful to be able to open up live examples and play around with them, but I hit save out of habit so now I won't mess up the original example :)

react/week-19/lesson.md
```diff
@@ -332,57 +338,55 @@ Here we are using `Array.map` to turn an array of strings into an array of compo
...
-      )
+      return <li key={index}>{name}</li>;
     })}
   </ul>
-)
+);
 ```
 
-Here we have added a `key` prop the to element. A documentation page explaining in more depth is in the further reading section but basically the `key` prop has a special meaning in React because it is used internally to keep track of which element in the list is which.
+Here we have added a `key` property to the `li` element. A documentation page explaining in more depth is in the further reading section but basically the `key` prop has a special meaning in React because it is used internally to keep track of which element in the list is which.
```
alismith-dev: My feeling generally we refer to `key` as a "prop", so I think for consistency I'd prefer to keep that.

react/week-19/lesson.md
```diff
@@ -399,55 +403,53 @@ The convention is to name component files the exactly same as the component (inc
 
 ## Making an Argument for Props
 
-What's the problem with our `HelloMentor` component above? Hint: imagine what our boss might ask for with this small application. What could our boss ask for which would mean we would have to make changes to the code?
```
alismith-dev, line 402: I think this a good call to remove this from the syllabus text (I can get a bit wordy, sorry!).

However I have found that it is useful to contextualise examples in a "real world" example like a job. I've observed that our students don't really have a good idea of what a development job looks like on a daily basis. Maybe it's worth giving a story like this in the class?

alismith-dev (commented):
This is really great! Thanks @llh1! I've left a few comments/questions :)

alismith-dev:
> We taught students to write JS with ';' at the end of each line, and create-react-app also uses this convention, so exercises and examples have been updated to use this as well.

Sorry! I don't use semis in my work (and personal projects), so it's just habit on my part. I try to fix that when I see it as we do teach semis in London too. Thanks for fixing those.

In terms of the CodeSandbox links, we should probably work out something so that we can collectively edit them. It's a bit of a pain to have to update the links in the syllabus if we change 1 minor thing. In the meantime I can go through and update all of the existing examples to use semis, if you think that is preferable.

react/week-19/lesson.md
llh1: Right i understand, I think mentor can maybe just show to the class that actually JSX is just Javascript finally instead of having it as an exercise? What do you think? Could add it as a mentor note.

react/week-19/lesson.md
llh1: Yes it's what we tend to do in Glasgow CYF, we don't necessarily stick to the syllabus when explaining or demonstrating things, we add some examples or stories that we say to the class which are not part of the syllabus. I think the syllabus can be seen as a reference for the theory but after it's up to each mentor to use the right examples to let students understand a concept. Maybe this could be worth adding to the mentor notes instead?

alismith-dev (approved):
Thanks for making those changes!

As discussed on slack, I'll transfer my existing CodeSandboxes to a CYF team and make sure that they get correctly prettified.

react/week-19/lesson.md
alismith-dev: I think adding it as a mentor note is a good idea 👍

alismith-dev:
@llh1 I've now transferred all of the CodeSandbox examples to a CYF team and prettified them. You have have an invite to the CodeSandbox team

react/week-19/lesson.md
alismith-dev: I think adding to the mentor notes is good - I've found that having concrete examples is good reference for new mentors

alismith-dev (approved):
Great! Thanks for that @llh1!

PR #289, opened 2019-05-06
Update React class 3

llh1:


alismith-dev (approved):
LGTM! :shipit: Thanks again @llh1

PR #290, opened 2019-05-07
React class 3: replace refs with working with forms section

llh1:
- Add semi colons
- Remove refs
- Add working with forms section and use only controlled component examples
- Add more complex example with a CreateAccountForm component.

### DIFF
--- a/react/week-21/lesson.md
+++ b/react/week-21/lesson.md
@@ -327,100 +327,114 @@ render() {
 > 16. **(STRETCH GOAL)** Add some error handling which renders an error message
 > 17. **(STRETCH GOAL)** Explore the data returned from the API. See if you can show some more interesting Pokemon information in your app (hint: try `console.log`ging different data returned from the API)
 
-## Refs
+## Working with forms in React
 
-As we have seen, React manages the DOM for us. That means we generally don't have to worry about keeping track of DOM nodes, or manipulating them directly. However sometimes you need to be able to access a DOM node that React is managing. Some common use cases are managing browser focus and integrating with third party libraries like a jQuery plugin.
+Modern web applications often involve interacting with forms such as creating an account, adding a blog post or posting a comment. This would involve using inputs, buttons and various form elements and being able to get the values entered by users to do something with it (like display them on a page or send them in a POST request). So, how do we do this in React?
 
-We can do this with a *ref*. Let's look at an example that will change browser focus to an input when a button is clicked ([interactive example](https://codesandbox.io/s/yw510x1l81)):
+A popular pattern for building forms and collect user data is the *controlled component* pattern. A pattern is a repeated solution to a problem that is useful in multiple similar cases. Let's have a look at an example ([interactive example](https://codesandbox.io/s/4jq1yqy8kx)):
 
-```js
-class InputFocuser extends Component {
-  setInputRef = (input) => {
-    this.input = input;
-  };
+```
+class SimpleReminder extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      reminder: ""
+    };
+  }
 
-  focusInput = () => {
-    this.input.focus();
+  handleChange = event => {
+    this.setState({
+      reminder: event.target.value
+    });
   };
 
   render() {
     return (
       <div>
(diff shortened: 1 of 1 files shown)


react/week-21/lesson.md
```diff
@@ -327,100 +327,114 @@ render() {
...
+      email: "",
+      password: ""
     };
   }
 
-  handleChange = (event) => {
+  handleChange = event => {
     this.setState({
-      name: event.target.value
+      [event.target.name]: event.target.value
```
alismith-dev, line 386: Do you think it's worth commenting on this below? I suspect students will have forgotten about this (do we even teach it at all?)

react/week-21/lesson.md
```diff
@@ -327,100 +327,114 @@ render() {
...
+      </div>
     );
   }
 }
 ```
 
-Now we're controlling the `value` of the input by using the value from state. This means that we can only change the value by updating state. If you didn't call `this.setState()` in the `handleChange` method, then the input's value would never change and it would appear as if you couldn't type in the input!
-
-This pattern is useful if you need to keep track of what the user is typing in the input. We could transform the string before we set it with `this.setState()`, for example by calling `toUpperCase()` on the string.
+We now have three different inputs named `username`, `email` and `password`, and we keep each entered value in a state with the same name. The method `handleChange` is reused to keep track of each change of value. The trick here is to use the name of the input element to update the corresponding state. Finally, when the user clicks on the submit button, the `submit` method is called to process the values. They are currently just displayed in the console but you could imagine validating the format of these values and sending them in a POST request.
```
alismith-dev, line 436: Previously I considered the refs/forms section "bonus material" that would only be covered if we had time in class - that's why it's at the end here.

But now I feel like it's slightly more important, so might be worth adding a few exercises?

react/week-21/lesson.md
```diff
@@ -327,100 +327,114 @@ render() {
...
-  setInputRef = (inputEl) => {
-    this.inputRef = inputEl;
-  };
+In addition, instead of just saving the value of the input in the state, we could have also transformed the string before we set it with `this.setState()`, for example by calling `toUpperCase()` on the string.
 
-  handleSubmit = (event) => {
-    console.log(this.inputRef.value);
-    event.preventDefault(); // Prevents form submission
-  };
+Let's have a look at a more complex example where we want to build a form to let users enter information to create an personal account ([interactive example](https://codesandbox.io/s/m7p083zn6p)):
```
alismith-dev:
```suggestion
Let's have a look at a more complex example where we want to build a form to let users enter information to create a personal account ([interactive example](https://codesandbox.io/s/m7p083zn6p)):
```

react/week-21/lesson.md
```diff
@@ -327,100 +327,114 @@ render() {
 > 16. **(STRETCH GOAL)** Add some error handling which renders an error message
 > 17. **(STRETCH GOAL)** Explore the data returned from the API. See if you can show some more interesting Pokemon information in your app (hint: try `console.log`ging different data returned from the API)
 
-## Refs
+## Working with forms in React
```
alismith-dev, line 330: Could you update the table of contents at the top too

alismith-dev (commented):
Looking good, thanks @llh1 👍 Left a couple of comments

alismith-dev:
Oh yes, I forgot too - if we're removing refs then we'll also need to update the instructions in the [week 3 homework](https://github.com/CodeYourFuture/cyf-hotel-react#lesson-3)

react/week-21/lesson.md
llh1: Yeah you're right, probably good to explain a bit this syntax

react/week-21/lesson.md
llh1: I also think the form part with controlled component is important and really need to be seen in class, students are going to use it a lot for their final group projects (at least they did in Glasgow). But they never really used Refs for example and we never really pushed them in that direction either. I can add more exercises following this.

alismith-dev (approved):
LGTM! 👍 Thanks @llh1!

alismith-dev:
As discussed on slack, we can add exercises later 🙂

PR #291, opened 2019-05-11
Update lesson.md

anthonytranDev:
Include Key as additional content page item - within the contents section


alismith-dev (approved):
Good catch! Thanks 🙂

PR #292, opened 2019-05-11
React improvements

alismith-dev:
A few minor changes to the React module, based on feedback from teaching London class 5. Summary of the changes:

- Remove `constructor` from examples/exercises with state initialisation. We found that this only added confusion. We ended up having to talk about class inheritance for a while, which a) doesn't help much with React understanding and b) isn't really what React uses classes for - classes are treated more like mixins. (I haven't updated the CodeSandboxes yet, but I can do that easily once this is merged)
- Improve the unmounting example by showing how React handles returning `null` in render instead of confusingly named component (`IsShown`/`IsNotShown`)
- Tweak homework instructions
- Provide a few more further reading links

I'm aware that this may have some conflicts with @llh1's [PR](https://github.com/CodeYourFuture/syllabus/pull/290), but happy to get that merged first and can then deal with the conflicts.

### DIFF
--- a/react/week-19/homework.md
+++ b/react/week-19/homework.md
@@ -2,5 +2,5 @@ Complete all of the lesson 1 exercises in the [cyf-hotel-react](https://github.c
 
 ### Prepare for the next class
 
-1. Follow the official [React tutorial](https://reactjs.org/tutorial/tutorial.html)
+1. Follow the official [React tutorial](https://reactjs.org/tutorial/tutorial.html) (up to the "Completing the Game" section)
 2. Watch the [Beginner's Guide to ReactJS](https://egghead.io/courses/the-beginner-s-guide-to-reactjs) on Egghead.io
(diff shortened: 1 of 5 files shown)


llh1:
My only concern about removing completely constructor() is that students may see a lot of examples on the web including the official documentation on React which uses constructor and may be clueless about this if we never introduce it. If using the class syntax, I feel overall we see more often React components written using constructor than without.

alismith-dev:
> My only concern about removing completely constructor() is that students may see a lot of examples on the web

Right. I also have this concern (actually if I could control what students were looking at, then we'd teach class components first - less context switching).

I would say that we did "waste" about 20 - 30 mins of class in London talking about class inheritance, which imo is not particularly efficient use of time (especially with Hooks, and the need for classes going away).

Perhaps we could add a short section explaining what the state class property does, and explain it with the use of a `constructor`?

>  including the official documentation on React

That is a good point.

alismith-dev:
I'll also get this rebased

alismith-dev:
@llh1 Was discussing this with Neill from London earlier today. To push this over the line I'm happy to revert the `constructor` change if you're still against it. I'd like to try switching over to Hooks for our next class anyway, which makes classes somewhat redundant.

llh1:
@40thieves Yes if you don't mind, i'd prefer to keep constructor for now as i think it's still more straightforward to explain just one way of doing things aligned with the documentation available on the web. But let's keep the discussion open on this, I'll pay attention of the students' response to this in few weeks in Glasgow and I'll let you know about it.

alismith-dev:
@llh1 I've pushed revert for the `constructor` changes (and fixed a missing `super(props)` call I spotted). Are you happy with this approach?

llh1:
@40thieves That's great :) thanks!

llh1 (approved):

PR #298, opened 2019-06-16
Write out REPL before using it

jslopezgithub:


bonboh:
Now fixed in the master branch

PR #299, opened 2019-06-18
React 4: Add introduction to React Router

llh1:
Add optional React module (that we will do in Scotland) to introduce routing to students.

### DIFF
--- a/react/week-22/lesson.md
+++ b/react/week-22/lesson.md
@@ -31,7 +31,7 @@ const Users = ({ names }) => {
 
 const AppRouter = () => {
   return (
-    <Router>
+    <BrowserRouter>
       <div>
         <nav>
           <ul>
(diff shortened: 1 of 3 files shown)


react/week-22/lesson.md
```diff
@@ -0,0 +1,120 @@
...
+            <li>
+              <Link to="/about/">About</Link>
+            </li>
+            <li>
+              <Link to="/users/">Users</Link>
+            </li>
+          </ul>
+        </nav>
+
+        <Route path="/" exact component={Index} />
```
textbook, line 50: Is it worth introducing this with `<Switch>` - the exclusive behaviour is probably less likely to cause surprises down the line than the inclusive behaviour.

llh1: That's a fair point, I wasn't sure but I tried finally to follow what is introduced on the official website of React Router in the quick start and avoid to introduce too many things at the same time. I think students can probably try `Switch` as a homework or further reading exercise? What do you think?

llh1: @textbook I'm going to merge this for now but happy to discuss it again.

alismith-dev:
Thanks for the contribution @llh1! Looks great :)

I've wondered in the past whether we should have an "extras" section for content like this. I deliberately left it out of the main course because I felt that it wasn't crucial to know for the majority of our students. But as an optional tutorial I think it's great.

How are you approaching teaching this lesson? Will all of the class take this lesson? My general thought is that it introduces several concepts at once (e.g. `<BrowserRouter>` - and the concept of providers - and `<Route>` together), which I feel some might struggle with. However I think that if we target it at the more advanced members of the class only, then I think it's fine.

llh1:
Hi @40thieves, sorry I missed this message, we taught React Router to everyone here in Glasgow, we didn't spend too much time on it, just a quick introduction with exercise, so they know it exists and that they can use it if needed. We didn't deal with the complexity of React Router, and just mentioned that if they want to define some route, they need to wrap everything in a top component called BrowserRouter, which comes from React Router library. I think the practice went well and that they saw the point of using this.

alismith-dev:
Sounds good 👍

PR #300, opened 2019-06-19
Add (rough) learning objectives for HTML module

nbogie:
These are still rough, but the suggested structure should make it easier for teachers to improve them.

I've split these by week to make them more convenient for a teacher addressing a particular week.  

(An alternative would be to leave them as a single document for the module in the top-level folder).

### DIFF
--- a/html-css/index.md
+++ b/html-css/index.md
@@ -1,7 +1,7 @@
 # HTML & CSS
 
-| Lesson                          | Content                                                                        |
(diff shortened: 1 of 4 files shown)


alismith-dev (approved):
I must admit I'm not super familiar with the module, but this looks good to me. Thanks for putting it together. I'll let @LucyMac @NateWr have a say since they know the material better than I

alismith-dev:
I also took the liberty of rebasing your branch to resolve the merge conflicts

PR #308, opened 2019-07-21
Add Career Development page

bonboh:
I've added a new page in the syllabus to give students information and resources on CVs, interviews, etc.

### DIFF
--- a/others/career-development.md
+++ b/others/career-development.md
@@ -31,6 +31,7 @@ CV samples
 
 - <https://www.livecareer.co.uk/cv-search/r/front-end-web-developer-31839afbd6b640429d637cdda953561c>
 - <https://www.livecareer.co.uk/cv-search/r/front-end-developer-00fdd3a63d27437d899f94a70df420be>
+- <https://www.brookes.ac.uk/students/careers/students-and-graduates/cv-guide/mature-graduate-cv/>—useful reference for people with gaps in their CV
 
 ## Interviews
 
@@ -43,7 +44,7 @@ Technical interview questions with answers (intermediate-level)
 - <https://github.com/30-seconds/30-seconds-of-interviews>
 - <https://github.com/FAQGURU/FAQGURU>
 
-Questions to ask the company (for inspiration)
+Questions to ask the company
 
 - <https://github.com/Twipped/InterviewThis>—some of these questions are great, but some are not so good
 
@@ -54,6 +55,7 @@ They are very useful for employers to see how you code and also demonstrate your
 
 Inspiration for personal projects
 
+- <https://github.com/florinpop17/app-ideas>
 - <https://github.com/karan/Projects>
 
 Walkthroughs of projects
--- a/others/career-development.md
+++ b/others/career-development.md
@@ -1,0 +1,73 @@
+# Career Development
+
+## CV
+
+### Best practice
+
+- Keep it short and concise, 1-2 pages maximum
+- Put your details at the top. Only these details are necessary:
+  - Name
+  - Email
+  - Phone number
+  - Your city (not your address since this is personal information! If they need it they'll ask for it later)
+- Use the following sections, and only populate them with relevant details starting from the most recent first
+  1. Education
+  2. Employment
+     - Use action verbs for each bullet point (see: <https://www.indeed.com/career-advice/resumes-cover-letters/action-verbs-to-make-your-resume-stand-out>)
+     - Show your accomplishments, with numbers if you can (e.g. Improved landing page time to interactive by 50%)
+  3. Skills
+  4. Personal projects
+  5. Personal interests (optional)
+
+### Resources
+
+Guide to writing a CV
+
+- <https://www.freecodecamp.org/news/writing-a-killer-software-engineering-resume-b11c91ef699d/>
+
+CV sample
+
+- <https://careercup.com/resume>
+
+## Interviews
+
+Interview questions with answers
+
+- <https://github.com/FAQGURU/FAQGURU>
+- <https://github.com/30-seconds/30-seconds-of-interviews>
+- <https://www.wisebread.com/how-to-answer-23-of-the-most-common-interview-questions>
+
+Questions to ask the company
+
+- <https://github.com/Twipped/InterviewThis>
+
+Quizzes
+
+- <https://davidshariff.com/quiz/>
+
+## Personal projects
+
+Personal projects are a good way to build up your portfolio of code projects on Github.
+They are very useful for employers to see how you code and also demonstrate your enthusiasm.
+
+Inspiration for personal projects
+
+- <https://github.com/karan/Projects>
+
+Walkthroughs of projects
+
+- <https://github.com/tuvtran/project-based-learning>
+- <https://javascript30.com/>
+
+## Contributing to open source
+
+Easier open source issues for people starting with open source contribution
+
+- <https://github.com/MunGell/awesome-for-beginners>
+- <https://firstcontributions.github.io/>
+
+## Miscillaneous
+
+Company reviews
+
+- <https://www.glassdoor.com/>
--- a/others/career-development.md
+++ b/others/career-development.md
@@ -9,41 +9,43 @@
   - Name
   - Email
   - Phone number
-  - Your city (not your address since this is personal information! If they need it they'll ask for it later)
+  - Your city (not your full address since this is personal information! If they need it they'll ask for it later)
 - Use the following sections, and only populate them with relevant details starting from the most recent first
-  1. Education
-  2. Employment
+  1. Summary (optional)
+  2. Skills
+  3. Employment
      - Use action verbs for each bullet point (see: <https://www.indeed.com/career-advice/resumes-cover-letters/action-verbs-to-make-your-resume-stand-out>)
      - Show your accomplishments, with numbers if you can (e.g. Improved landing page time to interactive by 50%)
-  3. Skills
-  4. Personal projects
-  5. Personal interests (optional)
+  4. Education (this can go above Employment if the last thing you did was education and not work)
+  5. Personal projects
+  6. Interests (optional)
 
 ### Resources
 
-Guide to writing a CV
+CV guides
 
-- <https://www.freecodecamp.org/news/writing-a-killer-software-engineering-resume-b11c91ef699d/>
+- <https://nationalcareers.service.gov.uk/get-a-job/cv-sections>
+- <http://www.energizerecruitment.co.uk/software-developers-guide-writing-cv-2017/>
 
-CV sample
+CV samples
 
-- <https://careercup.com/resume>
+- <https://www.livecareer.co.uk/cv-search/r/front-end-web-developer-31839afbd6b640429d637cdda953561c>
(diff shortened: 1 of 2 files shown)


others/career-development.md
```diff
@@ -0,0 +1,73 @@
...
+
+### Best practice
+
+- Keep it short and concise, 1-2 pages maximum
+- Put your details at the top. Only these details are necessary:
+  - Name
+  - Email
+  - Phone number
+  - Your city (not your address since this is personal information! If they need it they'll ask for it later)
+- Use the following sections, and only populate them with relevant details starting from the most recent first
```
p-maxwell, line 13: I think we might be best to advise them to skip having separate educational, employment and project sections, and list items in chronological order, making it clear what is educational, what is work, etc. Normally skills are at the top, then employment, then education, but that presumes you're currently in employment and education was a while ago. Their top item chronologically will probably be a portfolio project we'll have them working on.

others/career-development.md
```diff
@@ -0,0 +1,73 @@
...
+     - Show your accomplishments, with numbers if you can (e.g. Improved landing page time to interactive by 50%)
+  3. Skills
+  4. Personal projects
+  5. Personal interests (optional)
+
+### Resources
+
+Guide to writing a CV
+
+- <https://www.freecodecamp.org/news/writing-a-killer-software-engineering-resume-b11c91ef699d/>
```
p-maxwell: This is actually an American-style résumé, but since the article author has just graduated college, they don't have enough experience to make it much distinct from a UK CV.

Much of the advice is inappropriate for a CV. UK CVs are two pages, not one. Education shouldn't in general be at the top for anyone with actual experience (I think this is questionable for a US résumé also, but the author is a new college graduate...). Education coursework certainly shouldn't be on there if you have a reasonable quantity of real world experience (again this is probably also the case in the US market).

He does have good advice on showcasing accomplishments, but so much of the rest is misleading I would prefer not to use this link.

others/career-development.md
```diff
@@ -0,0 +1,73 @@
...
+
+CV sample
+
+- <https://careercup.com/resume>
```
p-maxwell: Again this is a résumé, not a CV. Not mentioning which technologies you've been working with role by role in a CV is non-standard to the point I'd suggest it's a bad idea.

others/career-development.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Interviews
+
+Interview questions with answers
+
+- <https://github.com/FAQGURU/FAQGURU>
```
p-maxwell: Hmm, some of these are going to be a bit difficult for our students as there's a lot of computer science stuff in there e.g. https://github.com/FAQGURU/FAQGURU/blob/master/topics/en/javascript.md#being-told-that-an-unsorted-array-contains-n---1-of-n-consecutive-numbers-where-the-bounds-are-defined-find-the-missing-number-in-on-time. The concept is solid though, maybe we could select some interview questions they wouldn't find too hard to practice with?

others/career-development.md
```diff
@@ -0,0 +1,73 @@
...
+- <https://github.com/30-seconds/30-seconds-of-interviews>
```
p-maxwell: I think this seems like it will probably be useful - they won't know everything, but these seem like sensible questions to know the answer to.

others/career-development.md
```diff
@@ -0,0 +1,73 @@
...
+- <https://www.wisebread.com/how-to-answer-23-of-the-most-common-interview-questions>
```
p-maxwell, line 40: This seems fairly reasonable as well.

others/career-development.md
```diff
@@ -0,0 +1,73 @@
...
+
+Questions to ask the company
+
+- <https://github.com/Twipped/InterviewThis>
```
p-maxwell: The first few seem reasonable. By the time I hit "object calisthenics" and asking about their brace style though... I think it would be a decent list for someone experienced who could filter out the nonsense, and wanted ideas for the three or four important ones, but I'm concerned there's so much noise it's likely to be more misleading than helpful.

others/career-development.md
```diff
@@ -0,0 +1,73 @@
...
+
+Quizzes
+
+- <https://davidshariff.com/quiz/>
```
p-maxwell: Haven't tried this, and it's not really my area of expertise - is it sensible?

others/career-development.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Personal projects
```
p-maxwell, line 51: We're going to be working with them on portfolio pieces at some length, so we probably want to cover how they're going to be describing that on the CV rather than giving them advice on personal projects in general.

others/career-development.md
```diff
@@ -0,0 +1,73 @@
...
+Easier open source issues for people starting with open source contribution
+
+- <https://github.com/MunGell/awesome-for-beginners>
+- <https://firstcontributions.github.io/>
+
+## Miscillaneous
+
+Company reviews
+
+- <https://www.glassdoor.com/>
```
p-maxwell: Would definitely take Glassdoor with a pinch of salt for many/most UK companies - the sampling rate is too poor to give much useful data.

others/career-development.md
```diff
@@ -0,0 +1,73 @@
...
+Walkthroughs of projects
+
+- <https://github.com/tuvtran/project-based-learning>
+- <https://javascript30.com/>
+
+## Contributing to open source
+
+Easier open source issues for people starting with open source contribution
+
+- <https://github.com/MunGell/awesome-for-beginners>
```
p-maxwell, line 70: This is paydirt. This is really excellent. I haven't dug much into this but it already looks like it might be an amazing resource. Thanks, Coung, this is something that could maybe make a huge difference.

others/career-development.md
```diff
@@ -0,0 +1,73 @@
...
+- <https://firstcontributions.github.io/>
```
p-maxwell, line 71: Similarly looks good, possibly a lot of the same information but that's not really a problem.

others/career-development.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Miscillaneous
```
p-maxwell: s/Miscillaneous/Miscellaneous/

others/career-development.md
p-maxwell: Finding a good UK resource for this is surprisingly challenging. https://nationalcareers.service.gov.uk/get-a-job/cv-sections gives some general CV advice, but it's not at all tech-focused. http://www.energizerecruitment.co.uk/software-developers-guide-writing-cv-2017/ is decent but bare-bones (and like most such guides, assumes the standard degree -> work progression).

others/career-development.md
p-maxwell: https://www.livecareer.co.uk/cv-search/r/front-end-developer-00fdd3a63d27437d899f94a70df420be is a pretty decent looking front end dev CV. I'd put skills at the top to make recruiter life easier. The bullet points could use a little work sadly. They're fine but unexciting - mercifully there are no typos. There's the usual issue for our students that it's the generic degree->work flow.

https://www.livecareer.co.uk/cv-search/r/front-end-web-developer-31839afbd6b640429d637cdda953561c is interesting because it's someone self-taught who shows off their portfolio in lieu of qualifications, though they only have one role to display and the formatting is a bit meh generally.

others/career-development.md
bonboh: Yeah on my CV I have Skills at the top but since I wasn't sure how conventional this was I put it in the bottom half since that was what was suggested in the Killer Resume link below. I'll put Skills at the top.

I also put Education above Employment against convention because I once read if you just graduated then it makes more sense to have Education first since that's what you most recently did and you don't have anything relevant to show in the Employment section. I'll switch this round.

I haven't heard of CVs without separate Education, Employment and Project sections. Is this format acceptable for employers?

others/career-development.md
bonboh: Thanks, I've updated the CV resource links

others/career-development.md
bonboh: I'll update the sample links with yours

others/career-development.md
bonboh: Yeah, my idea for this Career Development page is that it can also serve as a reference for graduates who have been working for a while. I'll add a note that these interview questions are more intermediate-level. If I find any beginner-level resources I'll add them. Would also be a good idea to list beginner-level questions from these resources.

others/career-development.md
bonboh: Okay I'll take out Glassdoor

others/career-development.md
bonboh: Whoops, I need to add a dictionary to my IDE

others/career-development.md
p-maxwell: So - this is a good question, and I'm mostly basing it off my own expectations since I've had to read a reasonably large number of CVs, and been behind the table for probably somewhere between a couple of dozen and thirty interviews, mostly as the main technical interviewer, and I'm trying to think what would annoy me least about a non-standard CV. Nick might have a lot more experience than me so I'd like to get his input and I'm currently looking to try to find articles on career change CVs and mature degree CVs to see if there's more general advice.

The basic problem is that for most people in the world of work, the employment and education sections are laid out entirely chronologically. In the case of our students, they won't be. If we put employment first, then the whole first page is maybe not relevant, if we put education first, then we will have two chunks of education that may be separated by a big gap. That might be okay, and we might be sensibly advised to have the sections and order them Projects - Education - Employment. I have found it really annoying having to hunt about different places on the CV to determine if there were gaps, so we need to be a little careful there I think.

p-maxwell: https://www.brookes.ac.uk/students/careers/students-and-graduates/cv-guide/mature-graduate-cv/ has a linked sample CV which looks interesting. It takes the Skills - Qualifications - Employment route, and I don't think the education gap would be too distracting here if it weren't for another iffy decision. While summarising work experience is reasonable, having it split from the rest of the employment section by volunteering is a bit more problematic. It took me quite a few seconds to chain together the GCSEs finishing 1989 into the summary starting 1990 to the Nursing Auxiliary role starting 1993. However, I like that the "career break" for kids is explained in employment, this may be relevant for some students.

There's a gap from 2001-2009 which is partly explained by volunteering one week per year 2002-2004, but is otherwise pretty weird, and would really benefit from an item similar to the career break item explaining why. Career gaps aren't necessarily deal-breakers, but as an interviewer you want to find out what happened - was your candidate in prison?

bonboh: Yeah having a CV laid out chronologically and ungrouped has the obvious benefit of avoiding jumping around to identify a sequential timeline. It would be good to get more input from others like Nick.

That mature CV is quite interesting. I've added it to the Career Development page.

bonboh:
@Peter-CYF does the Career Development page now look good to merge into the syllabus?

alismith-dev:
This is great! I'd love to hear some feedback from Scotland and we could get it merged to master

bonboh:
I already took the liberty of putting it in master so that other cities can benefit from it :) I just noticed that I need to add a note that the CV section is UK-focused.

Our students will be using the page soon and our plan is to tweak the page as we collect feedback from them.

PR #333, opened 2020-02-18
changing var to const

aldeez988:


js-core-1/week-1/lesson.md
```diff
@@ -200,22 +200,20 @@ var difference = 10 - 2; // 8
...
 Number of students: 15
 Number of mentors: 8
-Total numnber of students and mentors: 23
+Total number of students and mentors: 23
 ```
 
-## Floats
-
-Numbers can be integers (whole numbers) or floats (numbers with a decimal).
+Numbers can be written with, or without decimals:
```
ismaelgt, line 206: Although `float` is not a type in JavaScript, I still prefer the previous wording as an introduction to this terminology.

sansaid: ~Will probably update in a later commit (I agree with you here btw @ismaelgt ). Will approve for now.~

I take back what I said. As a compromise between your and @aldeez988 approach, would it be best if we kept the old wording and retitled the section to "Numbers as decimals" and change the previous section from "Numbers" to "Numbers as integers"?

sansaid:
Will merge - comments will be resolved at a later commit

PR #340, opened 2020-02-24
London class6 week2 update

sansaid:
* First pass at updating week 2 syllabus to match learning objectives
* Updated exercises

Exercises definitely need improving but would like some suggestions on what we can change it to to make it more challenging. Also uncertain about the `for` loops section, since I'm not sure if we want to make the students feel the pain of for loops so we can "sell' array methods for easily.


js-core-1/week-2/lesson.md
```diff
@@ -237,262 +271,135 @@ function satisfiesRequirements(num) {
...
-> var arr = [0, 1, 2, 3, 4]
-undefined
-> arr.slice(0, 2)
-[0, 1]
-> ["a", "b", "c", "d"].slice(1, 2)
-['b']
-```
-
-### `.includes()`
+Loops go hand-in-hand with objects like arrays in JavaScript. A **loop**  is a way for you to go over each element in an array (or more generally, an **iterable** - we'll come to this later).
```
ismaelgt: I wouldn't mention the relationship between loops and arrays so explicitly. I think they need to know that loops are what they need to use for repetition. Khan Academy has a nice review:

https://www.khanacademy.org/computing/computer-programming/programming/looping/a/review-looping

js-core-1/week-2/lesson.md
```diff
@@ -237,262 +271,135 @@ function satisfiesRequirements(num) {
...
 
-```sh
-$ node
-> ["H", "e", "l", "l", "o"].join();
-'H,e,l,l,o'
-> ["H", "e", "l", "l", "o"].join("--");
-'H--e--l--l--o'
+for (i in fruits) {
+  console.log(i)
+}
```
ismaelgt: I think you are meant to use `for/of` here? `for/in` is to iterate through object properties. I wouldn't bother mentioning any of these constructions, to be honest (just `while` and `for`). Specially if we want to encourage students to use `map`, etc.

js-core-1/week-2/lesson.md
sansaid: Ah, good point. See, this is what I needed to review, because I wasn't sure in what context we wanted to teach them loops (eventually the exercises will reflect that as well). Happy for the wording to be updated :)

js-core-1/week-2/lesson.md
sansaid: I did test the code and worked, but I wasn't sure what benefits `for/of` provides over `for/in` with respect to arrays (I thought they behaved the same for arrays). Is there a difference?

ismaelgt: [for...in](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in) iterates over object properties, whereas [for...of](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of) is for iterable objects, like arrays. Because you're using `for...in` this logs the indexes (array keys) and not the values:

Screenshot 2020-02-25 at 08 54 07

sansaid: Ah, thank you @ismaelgt ! I should question my memory more often :p

js-core-1/week-2/lesson.md
```diff
@@ -237,262 +271,133 @@ function satisfiesRequirements(num) {
...
-var orderedNumbers = unorderedNumbers.sort();
+let students = ["Ahmed", "Maria", "Atanas", "Nahidul", "Jack"]
 
-console.log(orderedLetters); // logs [ 'b', 'f', 'g', 'v', 'z' ]
-console.log(unorderedLetters); // logs [ 'b', 'f', 'g', 'v', 'z' ]
+students[2] = "Bianca"
 
-console.log(orderedNumbers); // logs [ 1, 2, 4, 5, 8 ]
-console.log(unorderedNumbers); // logs [ 1, 2, 4, 5, 8 ]
+console.log(students) // ["Ahmed", "Maria", "Atanas", "Nahidul", "Jack"]
```
aldeez988: @sansaid  there is a typo here in the output example the element in index number `2` should have the value of `Bianca ` not `Atanas`

js-core-1/week-3/lesson.md
```diff
@@ -15,6 +15,190 @@
...
+undefined
+> name.toLowerCase()
+daniel
+```
+
+Arrays also have several methods that you can use.
+
+### `.sort()`
+
+_An array method that sorts the values in an array into ascending alphabetical or numerical order._
```
aldeez988, line 80: The sort array also sort in `descending order` for example :
```
const points = [40, 100, 1, 5, 25, 10];
points.sort(function(a, b){return b-a});
```

js-core-1/week-3/lesson.md
```diff
@@ -15,6 +15,190 @@
...
+
+```js
+var unorderedLetters = ["z", "v", "b", "f", "g"];
```
aldeez988, line 83: It's better if we replace all the `var` keywords with `let` or `const`

js-core-1/week-2/lesson.md
sansaid: Thanks Elamin! I'll change that now.

js-core-1/week-2/lesson.md
```diff
@@ -35,11 +30,29 @@ The following are all examples of expressions:
...
 
 We can take the value produced by an expression and assign it to a variable. That line of code would be called a statement.
 
+Expressions can also contain variables.
+
+```js
+let planet = "Earth"
+
+`Hello ${planet}` // returns Hello Earth
+```
```
jamesgorrie:
```suggestion
```js
function() {
  const planet = "Earth"
  return `Hello ${planet}` // returns Hello Earth
}
```
```

File: js-core-1/week-2/lesson.md [hunk 7]
```diff
@@ -53,58 +66,47 @@ sayGreeting(greeting); // action: calls the function `sayGreeting` with the para
...
-undefined
-> greeting
-'hello'
-> console.log(greeting)
-hello
-undefined
-> .exit
-$
-```
+You can run `node` by itself, which will open a _node console_, also called a _Read–Eval–Print Loop (REPL)_. This console allows you to run expressions in the console line by line and is a great way of testing bits of code before writing it in a script.
```
jamesgorrie:
```suggestion
You can run `node` by itself, which will open a _node console_, also called a [Read–Eval–Print Loop (REPL)](https://en.wikipedia.org/wiki/Read%E2%80%93eval%E2%80%93print_loop).

This console allows you to run expressions in the console line by line and is a great way of testing bits of code before writing it in a script.
```

js-core-1/week-2/lesson.md
```diff
@@ -35,11 +30,29 @@ The following are all examples of expressions:
...
 
 We can take the value produced by an expression and assign it to a variable. That line of code would be called a statement.
 
+Expressions can also contain variables.
+
+```js
+function() {
+  const planet = "Earth"
+  return `Hello ${planet}` // returns Hello Earth
+}
```
sansaid:
```suggestion
}
\`\`\`
```

sansaid, line 44: Commenting to remember to add closing backticks

js-core-1/week-2/lesson.md
```diff
@@ -53,58 +66,49 @@ sayGreeting(greeting); // action: calls the function `sayGreeting` with the para
...
+
+This console allows you to run expressions in the console line by line and is a great way of testing bits of code before writing it in a script.
 
-> Notice how when we execute an expression the value it produces is printed below it. When we execute a statement, we see `undefined` printed below. This is because statements don't produce values like expressions, they _do something_.
+### Exercise (5 mins)
 
-* Write some more expressions in the node console
-* Assign some expressions to variables
-* Check the value of the variables
+In your VS Code terminal, run the command `node` and run the following expressions. What are their outputs? Is there anything you didn't expect? (To exit the node REPL, you have to click Ctrl+D or Cmd+D on Mac)
```
jamesgorrie:
```suggestion
In your VS Code terminal, run the command `node` and run the following expressions. What are their outputs? Is there anything you didn't expect? (To exit the node REPL, you have to click <kbd>Ctrl+d</kbd> or <kbd>Cmd+D</kbd> on Mac)
```

js-core-1/week-2/lesson.md
jamesgorrie, line 44:
```suggestion
  return `Hello ${planet}` // returns Hello Earth
}
```
```

File: js-core-1/week-2/lesson.md [hunk 10]
```diff
@@ -237,262 +294,210 @@ function satisfiesRequirements(num) {
...
-  }
+let i = 0;
+while (i < 20) {
+  someFunction(i);
+  i++;
 }
 ```
 
-You can also write this using `else if`:
+It's important that the condition inside the parenthesis becomes false at some point - otherwise we'll have what's known as an infinite loop!
```
jamesgorrie, line 332:
```suggestion
It's important that the condition inside the `while` condition becomes false at some point - otherwise we'll have what's known as an infinite loop!
```

js-core-1/week-2/lesson.md
```diff
@@ -237,262 +294,210 @@ function satisfiesRequirements(num) {
...
+
+The `for` loop is similar to a while loop, but with a more specialized syntax. Programmers invented the for loop when they realized they were always doing the same three things: creating loop counter variables (like `i` above), incrementing them by some amount, and checking that they're less than a value. A `for` loop syntax has special places for each of those three things. Here's the same loop as the first while loop above, as a for loop:
```
jamesgorrie, line 335:
```suggestion

![For loop](https://user-images.githubusercontent.com/31692/75388870-9213a880-58dd-11ea-90e6-4e67eabf390b.png)
```

js-core-1/week-2/lesson.md
```diff
@@ -237,262 +294,210 @@ function satisfiesRequirements(num) {
...
```
jamesgorrie:
```suggestion
The `for` loop is similar to a while loop, but with a more specialized syntax. Programmers invented the for loop when they realized they were always doing the same three things: creating loop counter variables (like `i` above), incrementing them by some amount, and checking that they're less than a value.

The `for` loop syntax has special places for each of those three things. Here's the same loop as the first while loop above, as a for loop:
```

js-core-1/week-2/lesson.md
```diff
@@ -237,262 +294,210 @@ function satisfiesRequirements(num) {
...
+let grades = ["F", "D", "A"];
+let greetings = ["Hello, how are you?", "Hi! Nice to meet you!"];
 ```
 
-Did you notice how we use `[0]` to get the first value? In programming we count starting at zero.
-
-> The number inside of the brackets is called an **index**. Index just means the position of the item within the array.
-
-You can also **set** a value using bracket notation and an assignment operator (`=`).
+You can access elements in an array using the **index** of an element with **bracket notation** (remember: all arrays start at position 0! So to access the first element in an array, you need to access index 0!)
```
jamesgorrie: This is important, so just pulling it out.

```suggestion
You can access elements in an array using the **index** of an element with **bracket notation**

**🔔 Remember:** All arrays start at position 0! To access the first element in an array, you need to access index `0`, the second element at `1`, the fifth at `4` and so forth. This is called zero indexed arrays. There are some [very intense reasons for this](http://www.cs.utexas.edu/users/EWD/transcriptions/EWD08xx/EWD831.html), but most people just accept it and move on.
```

js-core-1/week-2/lesson.md
```diff
@@ -237,262 +294,210 @@ function satisfiesRequirements(num) {
...
 
 ```js
-var scores = [80, 41, 47];
+let students = ["Ahmed", "Maria", "Atanas", "Nahidul", "Jack"]
 
-scores[2] = 29; // Change the last score
-scores[3] = 51; // Add a new score
+students[0] // Ahmed
+students[3] // Nahidul
```
jamesgorrie: Making sure types are reflected in comment outputs

```suggestion
students[0] // "Ahmed"
students[3] // "Nahidul"
```

js-core-1/week-2/lesson.md
```diff
@@ -237,262 +294,210 @@ function satisfiesRequirements(num) {
...
-var orderedNumbers = unorderedNumbers.sort();
+let students = ["Ahmed", "Maria", "Atanas", "Nahidul", "Jack"]
 
-console.log(orderedLetters); // logs [ 'b', 'f', 'g', 'v', 'z' ]
-console.log(unorderedLetters); // logs [ 'b', 'f', 'g', 'v', 'z' ]
+students[2] = "Bianca"
 
-console.log(orderedNumbers); // logs [ 1, 2, 4, 5, 8 ]
-console.log(unorderedNumbers); // logs [ 1, 2, 4, 5, 8 ]
+console.log(students) // ["Ahmed", "Maria", "Bianca", "Nahidul", "Jack"]
```
jamesgorrie, line 398: It's good we bring this up, but as a bikeshedding comment, I would like to chat about when we teach good practise and move on (in this case not mutating arrays) - but potentially brining it up as a thing they might see, but shouldn 't do.

Going into an interview knowing that mutation can cause bugs and create issues would be a definite plus to someone going into an interview.

jamesgorrie: Not sure who to ping on this ☝️

jamesgorrie (approved):
I'd say it looks generally good, we can merge and maybe make changes in smaller PRs?

js-core-1/week-3/lesson.md
jamesgorrie: Just changed this.
We should actually be using `const` where possible I think. @ismaelgt @sansaid thoughts?

js-core-1/week-2/lesson.md
alismith-dev: Hmm, good question. I think it's best to play it by ear. If you feel confident about the students, your explanation and have enough time, then go for it. One thing to keep in mind is that you have the experience of the "real-world", and it's useful to talk about it, if the time is right.

I'd suggest trying to ground it in an analogy like "changing the contents" of the array, but name-checking mutation so that they know there is some jargon here. I think it's also important to discuss _why_ it's a best practice: something along the lines of if you mutate it, then you can't "trust" whats inside it?

Finally I'd be careful with words like "shouldn't", as students can interpret that very literally. I'd go with some wording like "avoid" or "most of the time it's a bad idea" (as It Depends™️).

js-core-1/week-2/lesson.md
ismaelgt: > condition inside the `while` condition

Sounds redundant?

PR #338, opened 2020-02-22
Git intro

rarmatei:
Recommend clicking "View File" when looking at the lessons, as it's a **much** better experience:

image

PR #344, opened 2020-03-03
Add Github Desktop Half-Day Module

ChrisOwen101:
This has been run on the London Branch with a good amount of success.

The aim of this module is to teach enough Github & Github Desktop in half a day to make it easier for the students to be able to submit homework. The aim is to teach a more comprehensive view of Git in the future.

The thinking behind this is that Git introduces a huge amount of intrinsic and extraneous load on the students because we have to teach Command Line, Git and Github. This removes 2/3rds of the complexity whilst retaining the same results - the students can now submit homework.


sansaid:
Hey @ChrisOwen101 , thanks for this! Just to highlight - the narrative in the rest of the syllabus reads as if the audience is the student, where as the narrative in this lesson reads as if it is a guide for teachers. It would be best if we maintain consistency so we don't confuse readers on who the syllabus should be for. Probably worth a discussion, though it would be more work to change the narrative of the syllabus elsewhere and it also has the added benefit that the narrative currently in use allows the syllabus to be used by everyone.

git/week-1/lesson.md
```diff
@@ -0,0 +1,107 @@
...
+## What is the problem?
+
+When programmers work together it can be hard to stay in sync. This can be demonstrated by:
+
+- Have one person write three words on a piece of paper
+- Give the paper to the teacher and the teacher makes a copy of it
+- Somebody else can request the paper from the teacher
+- Change a word on the paper and give it back to the teacher.
+
+The teacher is the single source of truth for the project. This is fine so long as both people don't make a change at the same time.
```
alismith-dev, line 22: Not sure it's clear (even to me) what the "single source of truth" means?

I think it's trying to imply that the teacher makes a consistent 3 words after the student changes.

git/week-1/lesson.md
```diff
@@ -0,0 +1,107 @@
...
+
+- Now both students change a word on their piece of paper
```
alismith-dev, line 24: The steps above imply that there's only 1 piece of paper? Does this mean that a clone is made when given to the other student?

git/week-1/lesson.md
```diff
@@ -0,0 +1,107 @@
...
+- They then try to give both pieces of paper to the teacher
+  - But the teacher doesn't know what to do! Which one is correct?
+
+This is why Git was invented!
```
alismith-dev, line 28: I think we need to make it clear that the 3 words in this context are equivalent to code.

git/week-1/lesson.md
```diff
@@ -0,0 +1,107 @@
...
+
+## What is Github?
+
+**Confirm that everyone has a Github account and that they've been added to the relevant Github Teams**
+
+1. Give an overview of Github including
+   - A Project Overview
+   - Issues
+   - Creating a repository
+   - Changing a file in a repository
```
alismith-dev, line 38: I agree the comments that @sansaid made here: without the context of the teacher this isn't super helpful. This makes it difficult for the students to read as revision and for teachers in other cities to give good explanations.

I think we should flesh this out to explain what the overview actually means.

git/week-1/lesson.md
```diff
@@ -0,0 +1,107 @@
...
+
+- Now both students change a word on their piece of paper
+- They then try to give both pieces of paper to the teacher
+  - But the teacher doesn't know what to do! Which one is correct?
+
+This is why Git was invented!
+
+## What is Github?
+
+**Confirm that everyone has a Github account and that they've been added to the relevant Github Teams**
```
alismith-dev, line 32: Perhaps this could be moved to instructor notes.

git/week-1/lesson.md
```diff
@@ -0,0 +1,107 @@
...
+
+1. Give an overview of Github including
+   - A Project Overview
+   - Issues
+   - Creating a repository
+   - Changing a file in a repository
+2. All students should now
+   - Create a repository called "Code Your Future". This should include the ReadMe.md.
```
alismith-dev, line 40: Couple of minor points:

- Github will convert the repo name into `code-your-future`, so it might be better to just use that in the instructions
- The point about the readme - should it be auto-created during the repo creation flow?

git/week-1/lesson.md
```diff
@@ -0,0 +1,107 @@
...
+   - Edit the ReadMe.md inside Github to include their name.
```
alismith-dev, line 41: Maybe we should include some screenshots of the steps above?

git/week-1/lesson.md
```diff
@@ -0,0 +1,107 @@
...
+
+## What is Github Desktop?
+
+Github desktop is a way to edit files on your computer instead of editing them online.
+
+Give a demonstration of
```
alismith-dev, line 47: Similar comments as above about including some kind of explanation/screenshots here.

git/week-1/lesson.md
```diff
@@ -0,0 +1,107 @@
...
+
+1. Cloning a repository
+2. Making a change to a file in the repository
+3. Committing the changes to our local repository
+4. Pushing our changes to Github
+
+The students should now repeat these steps on their projects
+
+## What about working in Teams?
```
alismith-dev, line 56: Do we actually get the students to work in teams doing their homework? I wonder if this should go later in the course? On the other hand, I do think this is an interesting exercise to show the difference between git and say Dropbox.

Also, minor side point: why is teams capitalised? Seems to imply that it's a product or something.

alismith-dev (commented):
Generally I agree with @sansaid's comments. However without much effort it could be fleshed out to be useful outside of the lesson (for revision).

The other thought I had was that it might be difficult for teachers without context (for example in different cities).

alismith-dev:
Also, apologies on the delay in getting back to this!

alismith-dev:
Also just flagging this: https://github.com/CodeYourFuture/syllabus/pull/338

ChrisOwen101:
@40thieves Completely forgot about this - I've got to merge it so it's more obvious for Birmingham but I'll definitely add your comments tomorrow

PR #345, opened 2020-03-04
JS1 Week 3 Updates

ismaelgt:


js-core-1/week-3/lesson.md
```diff
@@ -32,7 +28,7 @@ You can check this by starting a node console in your terminal.
 
 ```sh
 $ node
-> var arr = [1, 2, 3];
+> let arr = [1, 2, 3];
```
jamesgorrie, line 31: I'm still not really sure why we are using `let`?

js-core-1/week-3/lesson.md
```diff
@@ -1,13 +1,14 @@
...
-* [Array ForEach](#array-foreach)
+* [Array properties](#array-properties)
+* [Manipulating arrays](#manipulating-arrays)
+* [Array methods](#array-methods)
+* [Array map](#array-map)
+* [Array forEach](#array-foreach)
+* [Array filter](#array-find)
+* [Array find](#array-find)
 
 **Learning Objectives**
```
jamesgorrie, line 13:
```suggestion
## Learning Objectives
```

js-core-1/week-3/lesson.md
```diff
@@ -88,11 +84,11 @@ Arrays also have several methods that you can use.
...
 ```js
-var unorderedLetters = ["z", "v", "b", "f", "g"];
-var orderedLetters = unorderedLetters.sort();
+let unorderedLetters = ["z", "v", "b", "f", "g"];
+let orderedLetters = unorderedLetters.sort();
 
-var unorderedNumbers = [8, 5, 1, 4, 2];
-var orderedNumbers = unorderedNumbers.sort();
+let unorderedNumbers = [8, 5, 1, 4, 2];
+let orderedNumbers = unorderedNumbers.sort();
```
jamesgorrie, line 91: Is it worth mentioning with `sort` that it mutates the original array?

js-core-1/week-3/lesson.md
```diff
@@ -1,13 +1,14 @@
...
-* [Array Filter](#array-every)
-* [Array Map](#array-map)
-* [Array ForEach](#array-foreach)
+* [Array properties](#array-properties)
+* [Manipulating arrays](#manipulating-arrays)
+* [Array methods](#array-methods)
+* [Array map](#array-map)
+* [Array forEach](#array-foreach)
+* [Array filter](#array-find)
+* [Array find](#array-find)
```
jamesgorrie, line 11: Nice order!

Should have a small piece on higher order functions / functions as values / whatever FP name you want as an intro to explain - and then use the array props as exercises to solidify that knowledge?

jamesgorrie (approved):
Nice nice nice!

js-core-1/week-3/lesson.md
sansaid: I thought it was because it's the latest syntax and enforces more control over scope and

js-core-1/week-3/lesson.md
sansaid: That's a really good point - I think we should definitely mention or add a sentence or two around it

sansaid: @jamesgorrie , perhaps this can be exemplified with an exercise?

js-core-1/week-3/lesson.md
ismaelgt: @jamesgorrie I'm not sure if you are referring to using `const` and `let` vs `var`, or using `const` vs `let`. About the first one, we talked about starting using ES6+ features as soon as possible. About the second one, I could have used `const` all along even if we're mutating the array, since what needs to remain constant is just the reference to it. I just thought that the fact that we can mutate an array declared using `const` could be confusing for students. But now that I think about it, I think it's ok to learn that ASAP (plus I introduced this concept last week anyway).

PR #346, opened 2020-03-05
Adding brief section on callbacks - not sure if this is a good idea

sansaid:
* Adding section on callbacks - not sure if this is a good idea yet, but thought I'd put it in the syllabus to see how it "feels" with the rest. Thoughts?

### DIFF
--- a/js-core-1/week-3/lesson.md
+++ b/js-core-1/week-3/lesson.md
@@ -16,6 +16,7 @@
 By the end of this class, you should be able to answer these questions:
 
 * What is a callback and how do you use it?
+* What are anonymous functions?
 * How do you use array methods and properties to manipulate arrays?
 * How can you chain array methods?
 
--- a/js-core-1/week-3/lesson.md
+++ b/js-core-1/week-3/lesson.md
@@ -2,6 +2,7 @@
 
 # JavaScript Core I - 3
 
+* [Callbacks](#callbacks)
 * [Array properties](#array-properties)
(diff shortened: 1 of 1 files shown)


js-core-1/week-3/lesson.md
```diff
@@ -15,11 +16,49 @@
 By the end of this class, you should be able to answer these questions:
 
 * What is a callback and how do you use it?
+* What are anonymous functions?
 * How do you use array methods and properties to manipulate arrays?
 * How can you chain array methods?
 
 ---
 
+## Callbacks
```
jamesgorrie, line 25: @sansaid I think this is a great idea.

We could start with explaining functions as values, as this is integral - and a way to use functions as values is callbacks.

What do you think?

jamesgorrie:
I think this is a great idea, I think the order in which things are introduced is a bit odd in the syllabus. We don't need to go into much depth, but something like this at least give the student and anchor as to why passing functions about is OK and they will be practising it a lot outside of the theory as it's just a javascripty thing.

ismaelgt (commented):
There is already a subsection on Callbacks under Array map

sansaid:
Ah, that's an oversight! Thanks @ismaelgt ! Should we keep it where it is or move it to its own section? We can keep it as it is, but I was thinking about teaching it in the start to the students anyway. Thoughts? @aldeez988 , @jamesgorrie ?

aldeez988 (approved):

sansaid:
Closing PR as callbacks section already added

PR #360, opened 2020-03-20
Update Homework for HTML/CSS

ChrisOwen101:
I've more the homework into its own Homework Module so that its centralised.

I've also moved the instruction from the homework so it only exists in the Homework repo so it is more in line with how the JavaScript homework is organised. 

I've used the homeworks from London's HTML/CSS module just because I know they're up to date and they work. Let me know anything that is a problem on Slack.


alismith-dev (approved):
I'm in favour of this, so approved.

However, perhaps this needs to be discussed on a call before merging? I think it intersects with the discussion about splitting up the syllabus into sub-modules and also a discussion about the way forward for the HTML/CSS module.

Also, if we merged this now would it affect any current teaching (e.g. Manchester)?

PR #361, opened 2020-03-20
Add Homework to JS2 - Week 3

ChrisOwen101:
Looks like the link to this has been removed at some point. This a small change to get this re-added.

### DIFF
--- a/js-core-2/week-2/lesson.md
+++ b/js-core-2/week-2/lesson.md
@@ -4,7 +4,7 @@
 
 **What we will learn today?**
 
-* [JS In the Browser (The DOM)](#js-in-the-browser)
+- [JS In the Browser (The DOM)](#js-in-the-browser)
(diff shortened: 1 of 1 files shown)


js-core-2/week-2/lesson.md
```diff
@@ -104,27 +104,27 @@ While it's really easy to change styles directly on elements using the `style` p
...
+  titleElement.innerText = title;
+  inputBox.value = title;
 });
 ```
 
 The above waits for click on a button. When the button is clicked, it gets the input box element (`inputBox` variable).
 To get the entered text from it, we use the `value` property: `var title = inputBox.value`.
 
-
+{% include "./homework.md" %}
```
ChrisOwen101, line 130: This is the only really important change

makanti (approved):

PR #369, opened 2020-04-03
Delete content now on CYF wiki

bonboh:
This PR deletes content that is now on https://docs.codeyourfuture.io. We should have a single source of truth to avoid outdated information and duplicitous efforts. The syllabus should also only be concerned with the actual web development course.


ChrisOwen101 (approved):
💯 💯 💯

bonboh:
Woohoo! I'll go ahead and merge then 💪

PR #370, opened 2020-04-04
Section on the call stack added.

BexB1:


js-core-2/week-3/lesson.md
```diff
@@ -3,18 +3,32 @@
...
 
 ![](https://img.shields.io/badge/status-draft-darkred.svg)
 
+## The Callstack
+
+How does JavaScript 'know' what order its code should be run in?
+
+JavaScript is a single-threaded language, which means that normally it can handle one task at a time or a piece of code at a time. It orders what it needs to do using something called the `call stack`. 
+
+The Callstack is a data structure that works by the "Last in, First out" principle (LIFO) to store and run functions.
```
illicitonion:
```suggestion
The Callstack is a data structure that works by the "Last in, First out" principle (LIFO) to store and run functions. Whenever you call a function, it gets pushed onto the stack, and when the function returns, it is popped off of the callstack.

This is why when you get an error in Javascript, you may see multiple lines with line numbers in the error, like:

$ node my.js
/home/dwh/my.js:2
    console.log(message);
                ^

ReferenceError: message is not defined
    at logSomething (/home/dwh/my.js:2:17)
    at computeSomething (/home/dwh/my.js:6:5)
    at Object.<anonymous> (/home/dwh/my.js:9:1)
    at Module._compile (internal/modules/cjs/loader.js:689:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:700:10)
    at Module.load (internal/modules/cjs/loader.js:599:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:538:12)
    at Function.Module._load (internal/modules/cjs/loader.js:530:3)
    at Function.Module.runMain (internal/modules/cjs/loader.js:742:12)
    at startup (internal/bootstrap/node.js:266:19)

This error happened because of a problem in the `logSomething` function on line 2, which was called by the `doSomething` function on line 6, and so on. Each line represents one entry on the callstack.
```

illicitonion (commented):
tbh I'm not sure whether explaining the callstack will make things more or less clear to folks, but it may be handy to draw the parallel to the stack shown when errors are raised to help visualise it :)

js-core-2/week-3/lesson.md
makanti: I like your idea Daniel. I think it's worth using a screenshot of the above reference error instead of copying the code in syllabus? It will help them visualise and remember when they will see it happening (pretty soon that is :))

BexB1: I like it. I'll commit as is, perhaps we can add a screenshot later.

makanti (approved):

PR #372, opened 2020-04-05
Manchester/update js core 1

anthonytranDev:
## Reason Changes

Proposed improvements on JS Core 1 - Weeks 1 - 3

Thinking of updating and possibly rebasing  this branch once Manchester starts doing the JS Core 1 module.

All comments are welcomed

### DIFF
--- a/js-core-1/week-2/lesson.md
+++ b/js-core-1/week-2/lesson.md
@@ -19,7 +19,7 @@ By the end of this class, you should be able to answer these questions:
 * What are conditionals?
 * What are `for` and `while` loops (and what's the difference)?
 * How do you use expressions with conditionals?
-* What are comparitor operators and how do you use them?
+* What are comparator operators and how do you use them?
 * What are logical operators and how do you use them?
 * What functionality does the `assert` module provide?
 * What are arrays and how can you interact with them?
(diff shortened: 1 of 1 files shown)


js-core-1/week-2/lesson.md
```diff
@@ -525,4 +540,14 @@ for (let i = 0; i < 10; ++i) {
 }
 ```
 
+## Glossary
+
+- Assert: to determine whether something is true or not true, more precisely false
+- Duplicate: exact copies of something (e.g. two or more files, numbers, directory can be exactly the same)
+- Index: numbers that let you know an item's position inside an array
+- REPL: (Read-Eval-Print-Loop) an interactive way to execute code you write inside the console
+- Zero-based Index: an arrays starting at `0` and not at `1`
```
ChrisOwen101:
```suggestion
- Zero-based Index: an array starting at `0` and not at `1`
```

js-core-1/week-2/lesson.md
```diff
@@ -525,4 +540,14 @@ for (let i = 0; i < 10; ++i) {
 }
 ```
 
+## Glossary
+
+- Assert: to determine whether something is true or not true, more precisely false
```
ChrisOwen101:
```suggestion
- Assert: to determine whether something is `true` or not `true`, more precisely `false`
```

ChrisOwen101 (approved):

PR #376, opened 2020-04-13
Add tasks to assess understanding for JS3

louisechow:
This PR adds a number tasks and questions that can be used either throughout the lesson or in homework to assess understanding of the JS3 content.

**Design**
* Each task matches a learning objective of the lesson. But, I did not cover every single one.
* The tasks attached to the _lesson plans_ were designed to take roughly 5 mins to run during the class.
* Because they were designed to be quick to answer, they mostly follow one the following formats:
    - Multiple choice question
    - Finish the code sample
    - Short answer 
* The tasks attached to the _homework plans_ were ones that could possibly take longer than 5 minutes (and hence were not suitable for use during the lesson).

**How to use tasks in the classroom** 
- The teacher shares the task in Slack. If the task has an example beforehand the teacher will talk through the example first.
- Give the students 1 - 2 minutes to complete the task and post their answer on the Slack thread.
- Teacher can spend 3 minutes to pick certain students to explain their answers and how they came to that solution.

### DIFF
--- a/js-core-3/week-1/lesson.md
+++ b/js-core-3/week-1/lesson.md
@@ -41,6 +41,22 @@ The purpose of this class is to introduce to the student:
 - Aggregated API: one API as a wrapper for multiple APIs.
 - Web services API: punch of APIs working together to forma whole app.
 
+
+##### Check-in
+
+**Question (5 mins):**
+
+    Which of the following statements below about APIs is false?
+        A) Public APIs can be accessed by anyone on the Internet.
+        B) You must use Javascript to access an API.
+        C) APIs can control access to data or features of an application.
+        D) You can change data via an API.
+
+**Question (5 mins):**
+
+    Give an example of a company that uses an API to allow access to their data.
+
+
 #### Basic structure of REST API
 
 - Endpoint: https://api.example.com
@@ -55,6 +71,24 @@ The purpose of this class is to introduce to the student:
 - Query params:
 
 * https://api.example.com/v1/users?limit=10
+
+
+##### Check-in
+
+Here is an example of Spotify's REST endpoint to allow you to get the tracks from a playlist
+
+    GET https://api.spotify.com/v1/playlists/{playlist_id}/tracks
+
+**Task (5 mins):**
+
+    On Instagram, you can access a post via the following endpoint:
+
+        GET https://api.instagram.com/v1/media/{media-id}
+
+    1. Write the URL of the endpoint to get the list of comments on a given instagram post?
+    2. What HTTP verb will you use to make requests to this endpoint?
+
+
 
 ### Example
 
--- a/js-core-3/week-1/lesson.md
+++ b/js-core-3/week-1/lesson.md
@@ -47,10 +47,11 @@ The purpose of this class is to introduce to the student:
 **Question (5 mins):**
 
     Which of the following statements below about APIs is false?
-        A) Public APIs can be accessed by anyone on the Internet.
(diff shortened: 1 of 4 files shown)


js-core-3/week-1/lesson.md
```diff
@@ -56,6 +72,24 @@ The purpose of this class is to introduce to the student:
...
+
+Here is an example of Spotify's REST endpoint to allow you to get the tracks from a playlist
+
+    GET https://api.spotify.com/v1/playlists/{playlist_id}/tracks
+
+**Task (5 mins):**
+
+    On Instagram, you can access a post via the following endpoint:
+
+        GET https://api.instagram.com/v1/media/{media-id}
```
xdl: Should we remove the `GET` here to make q2 less obvious?

js-core-3/week-1/lesson.md
```diff
@@ -96,6 +130,35 @@ promiseToDoHomeWork
   });
 ```
 
+##### Check-in
+
+**Question (5 mins):**
+
+    Which of the following statements about Promises is false?
+
+    A) A promise starts executing as soon as it is invoked.
```
xdl: Maybe rephrase this to 'The function passed into a promise is executed as soon as the promise is created.'?

js-core-3/week-2/lesson.md
```diff
@@ -85,6 +85,26 @@ window.VERSION = "1.0.4";
...
+    function firstFunction() {
+      console.log(number);
+    }
+
+    function secondFunction() {
+      let a = 5;
+      return firstFunction();
+    }
+
+    secondFunction();
```
xdl, line 112: Maybe we can make this a bit trickier, something like:

```
    let number = 1;

    function firstFunction() {

      number += 1;
      console.log(number);
    }

    function secondFunction() {

      let number = 5;

      number += 1;
      console.log(number);

      firstFunction();
    }
    
    firstFunction();
    secondFunction();
```

js-core-3/week-2/lesson.md
```diff
@@ -175,6 +195,30 @@ counterFromTen.whenClicked();
...
+
+}
+
+let myCart = new ShoppingCart(); // Creates an empty shopping cart
+myCart.addItem();
+```
+
+The output of running your code should be:
+
+       You have 1 item in your cart
```
xdl: Oo great exercise idea! I wonder if it's a bit similar to the Counter example code. How about something like this (or is it too tricky?):

```
class ShoppingCart {
// Add your code here
}
let myCart = new ShoppingCart(); // Creates an empty shopping cart
myCart.addItem("toilet roll");
myCart.addItem("pasta");
myCart.addItem("eggs");

myCart.listItems();
// "Your shopping cart has 3 items: toilet roll, pasta, eggs"
```

xdl (commented):
I added a couple of suggestions but looks great overall Louise!

js-core-3/week-1/lesson.md
louisechow: Ahhh good catch! Must have accidentally left that in.

js-core-3/week-2/lesson.md
louisechow: I'm don't know if all the students would be able to complete this in 2-3 minutes. The more advanced students perhaps, but the weaker students I'm not so sure.

We can use your example, but I'd suggest putting it into the homework rather than during the lesson and we find a different way to quickly assess understanding for the lesson.

PR #393, opened 2020-04-19
Update React learning objectives

alismith-dev:
Updates learning objectives for the React module:

- Extracted an "overall" set of learning objectives that is higher level and so will hopefully survive further React API changes (e.g. suspense)
- Updated the existing "weekly" learning objectives to remove references to class components ahead of the migration to Hooks
  - Left part of the week 3 objectives as a TODO, since I'm working on scoping this out currently
- Moved unused class component learning objectives to an "archived" file for easier retrieval or usage in additional material for advanced students

Feedback welcome!

@LucyMac @nbogie @anthonytranDev

### DIFF
--- a/react/overall-learning-objectives.md
+++ b/react/overall-learning-objectives.md
@@ -1,0 +1,31 @@
+# Overall Learning Objectives
+
+By the end of the three lessons & homework, students should be able to build a React application:
+
+- Written with components
+  - Appropriately composed/nested
+  - Split across multiple files
+- Uses props to pass values between components
+  - Data like numbers/strings/objects/arrays
+  - And functions for handling state changes
+- Uses state to react to user interaction
+  - E.g. a button can be clicked to increment a counter
+- That fetches data from an API
+  - Loading, success and error states are shown
+- Containing a form
+  - Form inputs that can be filled out by the user
+  - Can be submitted, triggering an update in the app
+
+<!-- TODO: use this to make a rubric? -->
+
+## Out of scope
+
+These topics are considered out-of-scope:
+
+- Object-Orientated JavaScript (classes/inheritance)
+  - Details of `extends`/`super()` are not necessary to use React (they are an implementation detail of `Component` class), so don't go out of the way to teach this
+- Redux
+  - Want to focus on learning and practising the basics of React, before introducing new concepts
+  - Learning Redux will be easier once confident with React
+- CSS-in-JS
+- In depth functional programming
--- a/react/overall-learning-objectives.md
+++ b/react/overall-learning-objectives.md
@@ -1,4 +1,4 @@
-# Overall Learning Objectives
+# Overall Learning Objectives - React module
 
 By the end of the three lessons & homework, students should be able to build a React application:
 
@@ -7,7 +7,7 @@ By the end of the three lessons & homework, students should be able to build a R
   - Split across multiple files
 - Uses props to pass values between components
   - Data like numbers/strings/objects/arrays
-  - And functions for handling state changes
+  - Functions as event handlers
 - Uses state to react to user interaction
   - E.g. a button can be clicked to increment a counter
 - That fetches data from an API
@@ -17,6 +17,11 @@ By the end of the three lessons & homework, students should be able to build a R
   - Can be submitted, triggering an update in the app
 
 <!-- TODO: use this to make a rubric? -->
+
+For advanced students, some further objectives would be:
+
+- Use class components instead of function components/Hooks
+- Explain why React state is used over plain old JavaScript variables
 
 ## Out of scope
 
--- a/react/week-1/learning-objectives.md
+++ b/react/week-1/learning-objectives.md
@@ -24,8 +24,3 @@
 ### Deployment
 
 - Build and deploy a React application (to Netlify/Now/Heroku/etc)
-
-### Out of scope
-
-- CSS-in-JS
-- In depth functional programming
--- a/react/week-1/learning-objectives.md
+++ b/react/week-1/learning-objectives.md
@@ -18,7 +18,7 @@
 ### Passing props
 
 - Pass values to components as React props
-  - Receive props in a component (both function and class components)
+  - Receive props in a function component
   - Pass functions as props and call them in child components
 
 ### Deployment
--- a/react/week-2/learning-objectives.md
+++ b/react/week-2/learning-objectives.md
@@ -24,6 +24,8 @@ Show some data from an API in a component.
 
 - Can explain why a loading state is necessary when fetching data as the result is not available on the initial render
 - Be able to trigger `fetch()` using the `useEffect()` callback
+  - Be able to an empty array (`[]`) as the 2nd dependencies argument
+  - Understand that this will be explained further in week 3
 - Be able to store data from a successful API call in state
   - And be able to explain that this causes a re-render
 - Be able to catch an unsuccessful API call, store the error in state and render an error message
--- a/react/week-2/learning-objectives.md
+++ b/react/week-2/learning-objectives.md
@@ -1,16 +1,27 @@
 ## Learning Objectives - React Week 2
 
-### Class components
-
-- Convert a function component to a class component
-
 ### Event Handlers
 
+React to user interaction via events.
+
 - Be able to pass functions to event handlers
+  - Log to the console when a button is clicked
 
 ### State
 
-- Create a simple counter component, demonstrating knowledge of React state
-  - Be able to initialise & render state variables
-  - Be able to set state using basic object style and based on previous state
+Create a simple counter component, demonstrating knowledge of React state.
+
+- Be able to use state variables
+  - Initialise with a value
+  - Render state variables to the screen
+  - Update with a new value
 - Identify when to use props or state
+
+### Data fetching
+
+Show some data from an API in a component.
+
+- Understand that the component will be rendered before data is available and handle this appropriately using a loading state
+- Be able to trigger `fetch()` using the appropriate method
+- Be able to store data from a successful API call in state, and be able to explain that this causes a re-render
+- Be able to catch an unsuccessful API call, store the error in state and render an error message
--- a/react/week-2/learning-objectives.md
+++ b/react/week-2/learning-objectives.md
@@ -14,7 +14,3 @@
   - Be able to initialise & render state variables
   - Be able to set state using basic object style and based on previous state
 - Identify when to use props or state
-
-### Out of scope
-
-- JavaScript classes/inheritance
--- a/react/week-2/learning-objectives.md
+++ b/react/week-2/learning-objectives.md
@@ -12,16 +12,18 @@ React to user interaction via events.
 Create a simple counter component, demonstrating knowledge of React state.
 
 - Be able to use state variables
-  - Initialise with a value
-  - Render state variables to the screen
-  - Update with a new value
+  - Initialise with a value with `useState(initialValue)`
+  - Destructure state variable into `[state, setState]`
+  - Render state variables to the view
+  - Update with a new value with `setState(newValue)`
 - Identify when to use props or state
 
 ### Data fetching
 
 Show some data from an API in a component.
 
-- Understand that the component will be rendered before data is available and handle this appropriately using a loading state
-- Be able to trigger `fetch()` using the appropriate method
-- Be able to store data from a successful API call in state, and be able to explain that this causes a re-render
+- Can explain why a loading state is necessary when fetching data as the result is not available on the initial render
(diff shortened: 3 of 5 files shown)


react/week-3/learning-objectives.md
```diff
@@ -1,22 +1,11 @@
...
 
-- "Clean up" a component when it is unmounted
-
-### Data fetching
-
-- Show some data from an API in a component
-  - Understand that `render` will be called before data is available and handle this appropriately
-  - Use `componentDidMount` lifecycle method to trigger `fetch()`
-  - Be able to hold data, loading & error states and render correctly based on these
+TODO (needs some scoping out)
```
LucyMac: Could include:

- Update state when a prop changes (via prop watcher argument in useEffect hook)

This will most likely be linked to the form input point outlined below so not sure if it should be included in that section instead.

LucyMac (commented):
Looks good to me, thanks @40thieves .

Are we purposely not mentioning specifics like around hooks here? If these were the obvious / only way to achieve the objectives it wouldn't necessarily need to be explicit, but as we're intentionally bringing it in, maybe we should clarify that we will be teaching useState and useEffect? Just a thought.

alismith-dev:
> Are we purposely not mentioning specifics like around hooks here?

When writing the overall objectives I was indeed trying avoid references to explicit techniques. Hopefully soon, the React team will release the Suspense API more widely and we'll have to go through this process again (since it will potentially make data fetching even easier).

Therefore I was hoping to write the final outcome objectives in a way which would survive those API changes, and potentially act as guidance when/if those changes do come through.

> maybe we should clarify that we will be teaching useState and useEffect? 

I think mentioning it explicitly in the weekly objectives is probably a good idea.

I think part of the reason they ended up like this is that learning objectives should generally attempt to be somewhat objectively assess-able. So for example "understands `useState()`" is not as good as something like "initialises state via `useState(initialState)`".

But as you can see it's probably fine to be a bit more specific, so I'll make that change. Thanks for the feedback 🙂

react/week-3/learning-objectives.md
alismith-dev: Could you expand a bit on what you mean? Perhaps a quick code snippet? (Sorry I think my inexperience with Hooks is not helping here)

I was planning on including something explaining the 2nd dependencies argument to `useEffect()` in more detail. Basically something along the lines of [this section in the official docs](https://reactjs.org/docs/hooks-effect.html#tip-optimizing-performance-by-skipping-effects).

And I was also planning on adding something about unsubscribing in `useEffect()`.

alismith-dev: I've pushed my initial ideas to the branch now

react/overall-learning-objectives.md
```diff
@@ -0,0 +1,36 @@
...
+  - Data like numbers/strings/objects/arrays
+  - Functions as event handlers
+- Uses state to react to user interaction
+  - E.g. a button can be clicked to increment a counter
+- That fetches data from an API
+  - Loading, success and error states are shown
+- Containing a form
+  - Form inputs that can be filled out by the user
+  - Can be submitted, triggering an update in the app
+
```
anthonytranDev, line 18: **Line 18)**

May want to include differences between controlled and uncontrolled components - this may be a further reading thing in the end

react/week-1/learning-objectives.md
```diff
@@ -18,14 +18,9 @@
 ### Passing props
 
 - Pass values to components as React props
-  - Receive props in a component (both function and class components)
+  - Receive props in a function component
   - Pass functions as props and call them in child components
```
anthonytranDev, line 22: **Line 21)**

Change `function` to `functional`

react/week-2/learning-objectives.md
```diff
@@ -1,20 +1,31 @@
...
-- Create a simple counter component, demonstrating knowledge of React state
-  - Be able to initialise & render state variables
-  - Be able to set state using basic object style and based on previous state
+Create a simple counter component, demonstrating knowledge of React state.
+
+- Be able to use state variables
+  - Initialise with a value with `useState(initialValue)`
+  - Destructure state variable into `[state, setState]`
+  - Render state variables to the view
+  - Update with a new value with `setState(newValue)`
```
anthonytranDev, line 18: Line 18)

Suggestion: 
Possible confusions with users of old React use of variables using `setState` as this part of React core API, i.e. this.setState() is an actual thing, that is used in Class Component React API - https://reactjs.org/docs/react-component.html#setstate

May want to change this to `value` and `setValue` instead

anthonytranDev (commented):
**Tool Suggestion**

Include section in using the Chrome plugin `react-dev-tools`, this may be harder for other cities.

In Manchester, however using Google Chrome DevTools was taught on Day 1 of HTML/CSS. So for use,  it doesn't so much as of a step up - might be a good idea, or something that'll be hard to integrate.  Food for thought!

react/week-1/learning-objectives.md
alismith-dev: The official docs don't (as far as I can tell) call them functional components. [This section](https://reactjs.org/docs/components-and-props.html#function-and-class-components) explicitly calls them function components.

I vaguely remember there being discussion on Twitter that technically components aren't actually functional (in the functional programming sense of the word), so they were avoiding that language.

react/overall-learning-objectives.md
alismith-dev: Yeah I think it may end up being further reading there is already a fair amount in week 3. Previously I had mostly focused on uncontrolled components but the consensus was that controlled was the more common pattern so it made sense to focus on that.

react/week-2/learning-objectives.md
alismith-dev: Yeah that's a good point. I'll update that. Maybe `stateValue` and `setStateValue()`?

alismith-dev:
> Include section in using the Chrome plugin react-dev-tools

I did consider this, but was struggling to think where it would fit, so it might end up as further reading or as an option section at the end of the lesson (so can be cut for time).

>  Google Chrome DevTools was taught on Day 1 of HTML/CSS

I'm not sure if it's officially in the syllabus but certainly in London it is taught. I would expect most other cities would too.

react/week-3/learning-objectives.md
```diff
@@ -1,22 +1,23 @@
...
-### Data fetching
-
-- Show some data from an API in a component
-  - Understand that `render` will be called before data is available and handle this appropriately
-  - Use `componentDidMount` lifecycle method to trigger `fetch()`
-  - Be able to hold data, loading & error states and render correctly based on these
+- Can explain what the effect of unmounting a component
+- Cleaning up effects
+  - Recognise when an effect could cause a memory leak
+  - Be able to prevent an effect causing a memory leak using the unsubscribe function of `useEffect()`
```
frosas, line 8: What example of a clean-up do you have in mind?

react/week-1/learning-objectives.md
```diff
@@ -18,14 +18,9 @@
...
 
 ### Deployment
 
 - Build and deploy a React application (to Netlify/Now/Heroku/etc)
```
frosas, line 26: Is the plan to do the bare minimum to use React (eg https://kentcdodds.com/blog/super-simple-start-to-react), or perhaps something like create-react-app?

react/week-3/learning-objectives.md
alismith-dev: I have been playing around with updating the "clock" example that was used in the old content:

- [Starting point](https://codesandbox.io/s/toggleable-clock-with-hooks-starting-point-gtqif?file=/src/Clock.js)
- [Mid point](https://codesandbox.io/s/toggleable-clock-with-hooks-mid-point-xqjn1?file=/src/Clock.js)
- [Final version](https://codesandbox.io/s/toggleable-clock-with-hooks-final-9ir6w?file=/src/Clock.js)

The idea being that the clock in the mid-point has a memory leak - check the console, when `Time` is unmounted - that they would have to fix by adding an unsubscribe function.

I'm vaguely aware that timers and `useEffect` can be a bit awkward, but I think for this example it works ok. And I think demonstrating a (minor) memory leak is somewhat useful too.

alismith-dev:
Weirdly I have an email notification where @frosas asked this:

> Is the plan to do the bare minimum to use React (eg https://kentcdodds.com/blog/super-simple-start-to-react), or perhaps something like create-react-app?

But I can't find that comment any more? (Hope you didn't delete it!)

Anyway to answer the question, we have historically gone with create-react-app. I wasn't aware of Kent's blog post so that's part of it. Skimming the post just now, it is quite similar to how we introduce React in the first lesson: https://syllabus.codeyourfuture.io/react/week-1/lesson.html#rendering-with-react.

There is some value in the eslint set up that comes with CRA. Certainly having the [Hooks linter rule](https://www.npmjs.com/package/eslint-plugin-react-hooks) will be quite helpful I think.

Having said that, we have had problems in the past with students laptops struggling to run CRA, I think because of the sheer amount of file watching/processing that happens in the background. I have wondered if it was possible to create a lighter CRA-alike because of this.

That would also allow us to set up [stop-runaway-react-effects](https://github.com/kentcdodds/stop-runaway-react-effects) by default, which I think I am leaning towards doing, as discussed here: https://codeyourfuture.slack.com/archives/C76GQDRH9/p1587308842005700

frosas:
> But I can't find that comment any more? (Hope you didn't delete it!)

I indeed deleted it as I could answer myself shortly after posting it. Sorry for the confusion, next time I'll just ask to ignore the comment 😅

> Anyway to answer the question, we have historically gone with create-react-app. I wasn't aware of Kent's blog post so that's part of it. Skimming the post just now, it is quite similar to how we introduce React in the first lesson: https://syllabus.codeyourfuture.io/react/week-1/lesson.html#rendering-with-react.

I really like the idea of not introducing any watch/build step from the very beginning as React brings quite a lot on itself 👍

> Having said that, we have had problems in the past with students laptops struggling to run CRA, I think because of the sheer amount of file watching/processing that happens in the background. I have wondered if it was possible to create a lighter CRA-alike because of this.

Maybe there's a CRA template for that but I don't think we can do much unless we cut features pretty common in any web app these days.

alismith-dev:
> I indeed deleted it as I could answer myself shortly after posting it. Sorry for the confusion, next time I'll just ask to ignore the comment 😅

No worries! It was still interesting to link to that Kent C. Dodds post!

> Maybe there's a CRA template for that but I don't think we can do much unless we cut features pretty common in any web app these days.

Indeed. We'll have to track how student laptops are holding up this time.

react/week-3/learning-objectives.md
LucyMac: Maybe this could be a stretch goal?

react/week-3/learning-objectives.md
```diff
@@ -1,22 +1,23 @@
...
+- Allow effects to update in response to prop changes
+  - Explain why a component with an effect dependent on props is broken with empty `useEffect()` dependencies (`[]`)
+  - Be able to fix a component with an effect dependent on props using the `useEffect()` dependencies
```
LucyMac, line 11: We're using the words "broken" and "fix" in the two points above - I haven't seen `useEffect` described that way before. Could we reword slightly? Suggestion below but by no means perfect.
- "explain how we can define that something (like fetching data for example) should happen only once when a component loads, by setting `useEffect`'s dependency to an empty array: `, []`".
- "explain how we can define that something (like a calculation for example) should happen each time a specific prop is updated, by setting `useEffect`'s dependency to that prop's name in square brackets: `, [specificProp]`".

LucyMac: We could potentially add something around the fact that there can be multiple useEffects in a component.
e.g. "can explain that `useEffect` can be used several times inside a component to respond to different triggers"

LucyMac: Sorry @40thieves I've just realised we already include the `[]` option in week 2's data fetching, makes more sense now! 🤦‍♀️

alismith-dev: The point I was attempting to make was that students know how to use `useEffect` (the double "use" really bothers me 😂) with no deps as you say. Then I was going to show them a situation where this doesn't work (i.e. is "broken"), discuss why it doesn't work and finally show them how to "fix" it.

But I agree that "broken" and "fix" aren't the best words to use here. I'll go with something like your wording, thanks 👍

react/week-3/learning-objectives.md
alismith-dev: I've written it up as part of my (upcoming) PR, it's a bit hard to tell how much time it will take to work through. My sense is that it won't take too long, but we can revisit this discussion when I open my PR?

PR #394, opened 2020-04-21
Tweaks to wording

illicitonion:


ChrisOwen101:
Thanks @illicitonion !

I've added these manually since we're still using the `javascript-core3` branch as the base and then merging into London (this is so we reduce conflicts when merging into master).

I'll close this PR but your changes are in there :)

PR #395, opened 2020-04-22
Rename Week Numbers to 1/2/3

ChrisOwen101:
In modules we usually name the Week 1, 2 or 3 so this change reflects this in the navigation.


alismith-dev (approved):
Thanks, been meaning to do this for a while. Doesn't reflect the actual length of the course either.

PR #401, opened 2020-04-26
Fetch error catching

xdl:
# What does this change?

Module: JS3
Week(s): Week 2

## Description

Adds in status codes and dealing with them with Fetch

@ChrisOwen101


alismith-dev:
Might want to mention [`response.ok`](https://developer.mozilla.org/en-US/docs/Web/API/Response/ok) too

dmitrigrabov (approved):

PR #405, opened 2020-04-27
Update React Lesson Week 2 - useState

anthonytranDev:
## Issues with Writing/Editing current docs

Wasn't clear on where I can include the `useState` section - this was my best guess.

I want to request a great deal more guidance here, if you believe the GitHub review tools aren't enough

Thank Goodness! The React team had easy to follow docs
https://reactjs.org/docs/hooks-overview.html

### DIFF
--- a/react/week-2/lesson.md
+++ b/react/week-2/lesson.md
@@ -4,11 +4,22 @@
 
 **What will we learn today?**
 
-- [Recap](#recap)
-- [Class Components](#class-components)
-- [Passing Functions as Props](#passing-functions-as-props)
-- [Reacting to Changes](#reacting-to-changes)
-- [State](#state)
+- [React 2](#react-2)
+  - [Recap](#recap)
+  - [Class Components](#class-components)
+    - [Class Methods](#class-methods)
+  - [Passing Functions as Props](#passing-functions-as-props)
+  - [Reacting to Changes](#reacting-to-changes)
+  - [State](#state)
+    - [When do you use Props or State?](#when-do-you-use-props-or-state)
+    - [Container components](#container-components)
+  - [React Hooks](#react-hooks)
+    - [What are hooks?](#what-are-hooks)
+    - [Why React hooks, over Class component state and methods?](#why-react-hooks-over-class-component-state-and-methods)
+    - [Counter Example - with hooks](#counter-example---with-hooks)
+    - [Setting multiple states](#setting-multiple-states)
+  - [Further Reading](#further-reading)
+- [Homework](#homework)
 
 ## Recap
 
@@ -428,6 +439,66 @@ In real world applications, the things we want to remember in state follow the *
 To help us cleanly split up code that performs business logic from code that shows the user interface we split components into *presentational* and *container* components. Often we have components that don't do anything except manage state according to the business rules and render the right presentational components. On the other hand, we often have components that don't change any state, and just render using the provided props.
 
 Container components usually have some state and handler methods. Because of this they must use the `class` syntax. Presentational components on the other hand don't require the more verbose syntax. Instead they usually use the functional syntax.
+
+## React Hooks
+
+React is continously updated with features all the time, one that stands out the most are **React Hooks**, released in React v16.8.
+
+### What are hooks?
+
+**Hooks** are functions that enable you manipulate the React state (and other lifecycle features, which you will be learning next week).
+
+### Why React hooks, over Class component state and methods?
+
+- Code is less complex - being able to make your code more readable in the long term, make it more maintainable
+- Hooks are reusable - unlike methods that scoped within in a class components, you are able to create your own and reuse them elsewhere you in the code. Keeping your code DRY (Don't Repeat Yourself)
+
+
+### Counter Example - with hooks
+
+```js
+import React, { useState } from 'react';
+
+function Counter() {
+  // Declare a new state variable, which we'll call "count"
+  const [count, setCount] = useState(0);
+
+  return (
+    <div>
+      <p>You clicked {count} times</p>
+      <button onClick={() => setCount(count + 1)}>
+        Click me
+      </button>
+    </div>
+  );
+}
+```
+
+Here, `useState` is a Hook. We call it inside a function component to add some local state to it. React will preserve this state between re-renders. `useState` returns a pair: the current state i.e. `count` and a function that lets you update it i.e. `setCount`.
+
+Note that variable `count` and function `setCount` come from destructured arrays (a JavaScript feature). The initial value of count is `0` and whenever we press on the button, it will fire the setCount function to increment by 1.
+
+### Setting multiple states
+
+```js
+function ExampleWithManyStates() {
+  // Declare multiple state variables!
+  const [age, setAge] = useState(0);
+  const [fruit, setFruit] = useState('Oranges');
+  
+  return (
+    <div>
+      <p> Hi! I am Alex </p>
+      <p> {`I am ${age} years old`} </p>
+      <p> {`I prefer to eat ${fruit}`} </p>      
+      <button onClick={() => setAge(age + 1)}>Add 1 year</button>
+      <button onClick={() => setFruit('Apples')}>Change to Apples</button>
+    </div>
+  )
+}
+```
+
+Notice that as before the initial state to a value inside `useState` in this case we have `0` and `Oranges`, with value `0` has the same behaviour as the counter example before. With value `Oranges`, it can turned into the value `Apple`. Stick to using the same data type, as it is best practice.
 
 ## Further Reading
 


react/week-2/lesson.md
```diff
@@ -429,6 +440,66 @@ To help us cleanly split up code that performs business logic from code that sho
 
 Container components usually have some state and handler methods. Because of this they must use the `class` syntax. Presentational components on the other hand don't require the more verbose syntax. Instead they usually use the functional syntax.
 
+## React Hooks
```
alismith-dev, line 443: To address the question about where to put to this, my plan was basically to move all of the existing content that uses classes to an "archive" or optional follow up lesson that is focused on Hooks.

Then we can put our new Hooks content in place of the old state lesson. I've been framing it as " teaching Hooks-first".

This also frees up time where we compare how state in classes and Hooks works (or at least moves it to an optional follow up that is focused on classes).

react/week-2/lesson.md
```diff
@@ -429,6 +440,66 @@ To help us cleanly split up code that performs business logic from code that sho
...
+
+React is continously updated with features all the time, one that stands out the most are **React Hooks**, released in React v16.8.
```
alismith-dev, line 445: I'm not really sure that our students care that much about the version history of React, since they're brand new to it. I wonder if this is just a distraction, so could be removed?

react/week-2/lesson.md
```diff
@@ -429,6 +440,66 @@ To help us cleanly split up code that performs business logic from code that sho
...
+
+### What are hooks?
+
+**Hooks** are functions that enable you manipulate the React state (and other lifecycle features, which you will be learning next week).
+
+### Why React hooks, over Class component state and methods?
```
alismith-dev, line 451: As discussed above, I think we should be treating this lesson as "Hooks-first". Meaning that comparing Hooks with classes would left to the follow-up classes lesson.

react/week-2/lesson.md
```diff
@@ -429,6 +440,66 @@ To help us cleanly split up code that performs business logic from code that sho
...
+
+- Code is less complex - being able to make your code more readable in the long term, make it more maintainable
+- Hooks are reusable - unlike methods that scoped within in a class components, you are able to create your own and reuse them elsewhere you in the code. Keeping your code DRY (Don't Repeat Yourself)
+
+
```
alismith-dev, line 456: What are your thoughts on adding some wording here about why the concept of state is necessary? Mostly just pointing out that the React apps we've shown so far are somewhat "boring" and mostly could be achieved using HTML.

In general I think our lessons are pretty good at showing the "how", but we often miss the "why".

The old state section has a paragraph about making some more interesting apps that _react to user interaction_: [see here](https://syllabus.codeyourfuture.io/react/week-2/lesson.html#reacting-to-changes).

alismith-dev: Also, please see my (extremely long, sorry!) comment about the aforementioned linked section here: https://codeyourfuture.slack.com/archives/C76GQDRH9/p1588106123031000

react/week-2/lesson.md
```diff
@@ -429,6 +440,66 @@ To help us cleanly split up code that performs business logic from code that sho
...
+import React, { useState } from 'react';
+
+function Counter() {
+  // Declare a new state variable, which we'll call "count"
+  const [count, setCount] = useState(0);
+
+  return (
+    <div>
+      <p>You clicked {count} times</p>
+      <button onClick={() => setCount(count + 1)}>
```
alismith-dev, line 469: In most of the examples I was creating, I pulled event handlers out into separate functions (within the scope of the component). I was thinking that this made it clearer.

But I don't really have anything to back that up. And I note that the official docs start with inlined functions too: https://reactjs.org/docs/hooks-reference.html#usestate

alismith-dev: I wonder if actually we could even split this up into 2 examples:

One that just shows the `count` variable being used (but is obviously broken), something like this

```js
function Counter() {
  const [count, setCount] = useState(0)
  return <p>You clicked {count} times</p>
}
```

And one that shows updating the count - i.e. this existing snippet.

This would allow us to discuss the array destructuring part before going into how the state is updated.

react/week-2/lesson.md
```diff
@@ -429,6 +440,66 @@ To help us cleanly split up code that performs business logic from code that sho
...
+        Click me
+      </button>
+    </div>
+  );
+}
+```
+
+Here, `useState` is a Hook. We call it inside a function component to add some local state to it. React will preserve this state between re-renders. `useState` returns a pair: the current state i.e. `count` and a function that lets you update it i.e. `setCount`.
```
alismith-dev, line 477: I wonder how quickly students will understand the "magic" aspect of `useState` preserving values between renders.

My feeling is that this may take a bit of explanation, but this could be a flavour of "expert's blindspot" in that I'm seeing a problem that most students wouldn't see.

react/week-2/lesson.md
```diff
@@ -429,6 +440,66 @@ To help us cleanly split up code that performs business logic from code that sho
...
+      <p> {`I prefer to eat ${fruit}`} </p>      
+      <button onClick={() => setAge(age + 1)}>Add 1 year</button>
+      <button onClick={() => setFruit('Apples')}>Change to Apples</button>
+    </div>
+  )
+}
+```
+
+Notice that as before the initial state to a value inside `useState` in this case we have `0` and `Oranges`, with value `0` has the same behaviour as the counter example before. With value `Oranges`, it can turned into the value `Apple`. Stick to using the same data type, as it is best practice.
+
```
alismith-dev, line 502: I wonder if we should keep the idea from the old content of demonstrating that component state is instanced (i.e. multiple `Counter`s have separate state): 

```js
function App() {
  return (
    <div>
      <Counter />
      <Counter />
      <Counter />
    </div>
  );
};
```

In my experience, this helps with "thinking in components".

react/week-2/lesson.md
```diff
@@ -429,6 +440,66 @@ To help us cleanly split up code that performs business logic from code that sho
...
+  const [age, setAge] = useState(0);
+  const [fruit, setFruit] = useState('Oranges');
+  
+  return (
+    <div>
+      <p> Hi! I am Alex </p>
+      <p> {`I am ${age} years old`} </p>
+      <p> {`I prefer to eat ${fruit}`} </p>      
+      <button onClick={() => setAge(age + 1)}>Add 1 year</button>
+      <button onClick={() => setFruit('Apples')}>Change to Apples</button>
```
alismith-dev, line 495: It might be nice to add a button that switches back to Oranges, otherwise there might be an assumption that it can only change 1 way.

react/week-2/lesson.md
alismith-dev: When I was looking at "advanced `useEffect`" was that `setStateValue` does **not** trigger a re-render if the value is considered "the same": https://reactjs.org/docs/hooks-reference.html#bailing-out-of-a-state-update.

This was important with `useEffect` as there are times when you can trigger infinite effect loops: https://codesandbox.io/s/trusting-cloud-zj858?file=/src/App.js.

I'm a bit unsure about if (or when) to teach this as it potentially gets into discussion about value equality in JS and how memory is stored which might be quite advanced.

alismith-dev (commented):
Thanks for taking this on @anthonytranDev! This is looking good 🙂 I've left some comments which I'm happy to discuss!

react/week-2/lesson.md
```diff
@@ -429,6 +440,66 @@ To help us cleanly split up code that performs business logic from code that sho
...
+function ExampleWithManyStates() {
+  // Declare multiple state variables!
+  const [age, setAge] = useState(0);
+  const [fruit, setFruit] = useState('Oranges');
+  
+  return (
+    <div>
+      <p> Hi! I am Alex </p>
+      <p> {`I am ${age} years old`} </p>
+      <p> {`I prefer to eat ${fruit}`} </p>
```
alismith-dev, line 493: Sorry for only just noticing this, but I would tend to write this like:

```suggestion
      <p> I am {age} years old </p>
      <p> I prefer to eat {fruit} </p> 
```

Was there a reason to use template strings?

react/week-2/lesson.md
LucyMac: I think it might be the "expert's blindspot" you mention 😉 I don't see the magic as a problem. We teach them how to use it and they can go and find out exactly how it works in their own time?

react/week-2/lesson.md
LucyMac: Agreed some introduction to State would be useful. It's a really unfamiliar word (or rather, an unfamiliar meaning for the word) that threw me when I was learning. 
Something like this maybe:

> React components display data - for example someone's name, some numbers, a picture, some information received from an API, etc.
> This data can be one of 2 things:
> a) Received from a parent component: these are known as props, which you learned last week.
> or
> b) Created and hosted inside a component: this is what we call "state". The component that 'owns' the state can update it, manipulate it, and also pass it down to its own child components as props.

react/week-2/lesson.md
alismith-dev: Just realised that first paragraph doesn't make any sense 🤦 I meant follow up lesson focused on classes.

react/week-2/lesson.md
alismith-dev: Yes sorry to be clear, this was more of an observation than a request for change: I think we'll find out pretty quickly during the lesson if students are confused here.

alismith-dev:
Apologies for the continual stream of comments here, but have you thought about what in-class exercises we might use? Would we just adapt the existing state exercises to be using `useState` instead?

alismith-dev:
As discussed on Slack, we're going to merge this now and make further changes as follow ups.

react/week-2/lesson.md
anthonytranDev: Thanks for the suggestions guys, I'll implement as much as I can and do a new PR for this. 😄

PR #409, opened 2020-05-02
DG update scope and this

dmitrigrabov:


js-core-3/week-3/lesson.md
```diff
@@ -115,7 +144,7 @@ function greetUser(name) {
 }
 ```
 
-But in the same way countries have different laws, functions have different `scopes`! And so it becomes possible to use the same variable name for different situations.
+But in the same way countries have own laws, functions have own `scopes`! And so it becomes possible to use the same variable name for different situations.
```
esmehm, line 147: *their own x2

js-core-3/week-3/lesson.md
```diff
@@ -388,7 +461,113 @@ myAddresBook.lookup("homer");
...
+
+```js
+function sayHey() {
+  console.log(this);
+}
+
+sayHey();
+```
+
+### methods as called as functions
```
esmehm, line 518: This doesn't quite make sense - maybe reword it slightly?

PR #411, opened 2020-05-03
Dg this exercises

dmitrigrabov:


js-core-3/week-3/lesson.md
```diff
@@ -562,9 +563,37 @@ sayHi(); // Hi, Mike
...
+```js
+// Exercise 1
+class Car {
+  goFaster() {
+    this.speed = 100; // what is the value of this code below is run
+  }
+}
+
+const ferrari = new Car();
+ferrari.goFaster();
```
xdl, line 579: Should we add what the expected output of calling this method is? Maybe it can call `goFaster` twice and the speed will be like 200?

```
...
constructor() {
  this.speed = 0;
}
goFaster() {
  this.speed = 100; // what is the value of this code below is run
  return this.speed;
}
...

ferrari.goFaster();
// 100
ferrari.goFaster();
// 200
```

dmitrigrabov: The main thing I wanted to illustrate is how the method is getting called rather than wider functionality.

js-core-3/week-3/lesson.md
```diff
@@ -562,9 +563,37 @@ sayHi(); // Hi, Mike
...
+class Car {
+  goFaster() {
+    this.speed = 100; // what is the value of this code below is run
+  }
+}
+
+const honda = new Car();
+const quicker = honda.goFaster;
+
+quicker();
```
xdl, line 591: And then this one should return `100`

PR #415, opened 2020-05-05
React use effect

LucyMac:
Adds useEffect for mounting to week 2 (for data fetches), and useEffect for updating to week 3.

May need some more fleshing out, and probably needs some exercises attached to these, but wanted to get initial feedback.

@40thieves it would also be super helpful to see the changes you've been making as there are mentions of classes throughout and most examples use class syntax.

### DIFF
--- a/react/week-2/lesson.md
+++ b/react/week-2/lesson.md
@@ -18,6 +18,8 @@
     - [Why React hooks, over Class component state and methods?](#why-react-hooks-over-class-component-state-and-methods)
     - [Counter Example - with hooks](#counter-example---with-hooks)
     - [Setting multiple states](#setting-multiple-states)
+  - [Fetching data in React](#fetching-data-in-react)
+    - [The `useEffect` hook](#the-useeffect-hook)
   - [Further Reading](#further-reading)
 - [Homework](#homework)
 
@@ -499,6 +501,57 @@ function ExampleWithManyStates() {
 ```
 
 Notice that as before the initial state to a value inside `useState` in this case we have `0` and `Oranges`, with value `0` has the same behaviour as the counter example before. With value `Oranges`, it can turned into the value `Apple`. Stick to using the same data type, as it is best practice.
+
+## Fetching data in React
+
+Often when you create a React app, you will want to get data from an API, and display it inside your components.
+How do we do this in React? Where does the API call go, and when should we trigger it?
+
+**Where:** Most likely at the top of the component tree, inside a 'container' component as explained above. You can then flow the data down into your child components as props.
+**When:** When the component is first loaded into the DOM. We call this 'mounting'.
+**How:** With a handy new hook called `useEffect`
+
+### The `useEffect` hook
+
+Just like `useState`, the `useEffect` hook is a special function that all function components have access to. This is the syntax to follow to fetch data on mount:
+
+```js
+useEffect(() => {
+  // Make your API call here!
+}, []); // Don't forget these empty square brackets
+```
+
+And here is a more complete example:
+
+```js
+import React, { useState, useEffect } from "react";
+
+const TvShowContainer = () => {
+  const [movies, setMovies] = useState([]);
+
+  useEffect(() => {
+    fetch("http://example.com/movies.json").then((response) => {
+      setMovies(response.json());
+    });
+  });
+
+  return (
+    <div>
+      {movies.length > 0 ? (
+        <div>There are {movies.length} to display</div>
+      ) : (
+        <p>Sorry, there are no movies to display</p>
+      )}
+    </div>
+  );
+};
+
+export default TvShowContainer;
+```
+
+In the code above, we're saying to React “When this component is mounted, call this API, and when you receive a response, save it inside of the 'movies' state”.
+
+This is a very common pattern which will come in very useful!
 
 ## Further Reading
 
--- a/react/week-2/lesson.md
+++ b/react/week-2/lesson.md
@@ -542,11 +542,7 @@ const MartianPhotoFetcher = () => {
 
   return (
     <div>
-      {movies.length > 0 ? (
-        <div>There are {movies.length} to display</div>
-      ) : (
-        <p>Sorry, there are no movies to display</p>
-      )}
+      // TODO: update this example
     </div>
   );
 };
--- a/react/week-2/lesson.md
+++ b/react/week-2/lesson.md
@@ -99,6 +99,7 @@ Here are the steps to follow to convert from a functional component into a class
 
 > **Exercise A**
 > Open the `pokedex` React application that you created last week
+>
 > 1. Convert the `Logo` component from a functional component into a class component
 > 2. Convert the `CaughtPokemon` component into a class component
 > 3. Convert the `BestPokemon` component into a class component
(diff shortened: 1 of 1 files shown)


react/week-2/lesson.md
```diff
@@ -429,6 +437,57 @@ To help us cleanly split up code that performs business logic from code that sho
...
+Often when you create a React app, you will want to get data from an API, and display it inside your components.
+How do we do this in React? Where does the API call go, and when should we trigger it?
+
+**Where:** Most likely at the top of the component tree, inside a 'container' component as explained above. You can then flow the data down into your child components as props.
+**When:** When the component is first loaded into the DOM. We call this 'mounting'.
+**How:** With a handy new hook called `useEffect`
+
+### The `useEffect` hook
+
+Just like `useState`, the `useEffect` hook is a special function that all function components have access to. This is the syntax to follow to fetch data on mount:
```
alismith-dev: > a special function that all function components have access to

I don't think this isn't quite true: you still need to import it. We should probably mention that in the `useState` section too now that I think about it.

react/week-2/lesson.md
```diff
@@ -429,6 +437,57 @@ To help us cleanly split up code that performs business logic from code that sho
...
+
+```js
+useEffect(() => {
+  // Make your API call here!
+}, []); // Don't forget these empty square brackets
```
alismith-dev: A couple of points.

I wonder if we should just say "empty array". "Square brackets" implies that it's somehow not an array.

My observation is that some students find it really hard to accept that they just have to type something in. (This was the case with `super(props)` in the old classes content). I think we mention that this will be explained in more depth in week 3.

react/week-2/lesson.md
```diff
@@ -429,6 +437,57 @@ To help us cleanly split up code that performs business logic from code that sho
...
+And here is a more complete example:
+
+```js
+import React, { useState, useEffect } from "react";
+
+const TvShowContainer = () => {
+  const [movies, setMovies] = useState([]);
+
+  useEffect(() => {
+    fetch("http://example.com/movies.json").then((response) => {
```
alismith-dev: I did quite like the NASA Rover images API we were using in week 3 before 😄 

My re-work of week 3 does still use that API, but I think it's fine to look at it again.

react/week-2/lesson.md
```diff
@@ -429,6 +437,57 @@ To help us cleanly split up code that performs business logic from code that sho
...
+
+Just like `useState`, the `useEffect` hook is a special function that all function components have access to. This is the syntax to follow to fetch data on mount:
+
+```js
+useEffect(() => {
+  // Make your API call here!
+}, []); // Don't forget these empty square brackets
+```
+
+And here is a more complete example:
```
alismith-dev, line 528: One of the things I really liked about the old content was that nearly all code snippets had "interactive examples" in CodeSandbox. I could open them up and play around with them to demonstrate a point.

Would it be possible copy this across to CodeSandbox?

react/week-2/lesson.md
```diff
@@ -429,6 +437,57 @@ To help us cleanly split up code that performs business logic from code that sho
...
+        <p>Sorry, there are no movies to display</p>
+      )}
+    </div>
+  );
+};
+
+export default TvShowContainer;
+```
+
+In the code above, we're saying to React “When this component is mounted, call this API, and when you receive a response, save it inside of the 'movies' state”.
```
alismith-dev: In my (upcoming) PR, I've explicitly written out all the steps to "play computer" line-by-line.

This plays nicely into the more advanced `useEffect`, where it is important to understand that `useEffect` is "queuing" an effect to be run after render.

react/week-3/lesson.md
```diff
@@ -42,41 +46,57 @@ class Counter extends Component {
 
 ## Unmounting
 
-So far we've looked at components that are always rendered in the browser. However (and this is often the case in large applications), we might want to control whether components are shown or not. Let's look at a Toggle component ([interactive example](https://codesandbox.io/s/xmo8oo514)):
+So far we've looked at components that are always rendered in the browser. However (and this is often the case in large applications), we might want to control whether components are shown or not.
```
alismith-dev: Everything up to the forms section in week 3 is stuff I've been working on. Perhaps we could move these changes/discussion to that PR when I open it (Soon™️)?

react/week-2/lesson.md
```diff
@@ -429,6 +437,57 @@ To help us cleanly split up code that performs business logic from code that sho
...
+      setMovies(response.json());
+    });
+  });
+
+  return (
+    <div>
+      {movies.length > 0 ? (
+        <div>There are {movies.length} to display</div>
+      ) : (
+        <p>Sorry, there are no movies to display</p>
```
alismith-dev: What do you think about teaching an explicit loading state?

Although I would probably separate that from the simple data fetching example here.

react/week-2/lesson.md
alismith-dev: This example is missing the `.then(res => res.json())` bit too. I think this is important to enable easy copy/paste.

alismith-dev (commented):
Looking good @LucyMac! Thanks for taking this on 👍 

Have you thought about the in-class exercises? Would we just adapt the exercises from the previous data-fetching content (the pokemon api)?

react/week-2/lesson.md
alismith-dev: Oh actually, looking through my changes I have actually covered some stuff on loading state. So perhaps not necessary.

react/week-2/lesson.md
LucyMac: Good point about importing. Maybe we can reword: 

> the `useEffect` hook is a special function that all function components can import and use as needed.

react/week-2/lesson.md
LucyMac: Yes absolutely  👍

react/week-2/lesson.md
```diff
@@ -429,6 +437,57 @@ To help us cleanly split up code that performs business logic from code that sho
 
 Container components usually have some state and handler methods. Because of this they must use the `class` syntax. Presentational components on the other hand don't require the more verbose syntax. Instead they usually use the functional syntax.
 
+## Fetching data in React
+
+Often when you create a React app, you will want to get data from an API, and display it inside your components.
+How do we do this in React? Where does the API call go, and when should we trigger it?
+
+**Where:** Most likely at the top of the component tree, inside a 'container' component as explained above. You can then flow the data down into your child components as props.
```
LucyMac: FYI I'd like to add an image here to illustrate the 'component tree'.

react/week-3/lesson.md
LucyMac: Yes sorry I got carried away 😬

react/week-2/lesson.md
alismith-dev: Wording looks good 👍 

I think we should show a quick example of importing. One of the things that trips students up is forgetting to import `React` in component files (so that JSX works)

react/week-2/lesson.md
alismith-dev: 👍 Just making a note that this is similar to the existing [Container Components](https://syllabus.codeyourfuture.io/react/week-2/lesson.html#container-components) section.

alismith-dev (approved):

react/week-2/lesson.md
```diff
@@ -429,12 +441,61 @@ To help us cleanly split up code that performs business logic from code that sho
...
+
+const MartianPhotoFetcher = () => {
+  const [marsPhotos, setMarsPhotos] = useState();
+
+  useEffect(() => {
+    fetch(`https://api.nasa.gov/mars-photos/api/v1/rovers/curiosity/photos?earth_date=2015-6-3&api_key=DEMO_KEY`)
+      .then(res => res.json())
+      .then(data => setMarsPhotos(data));
+    });
+  });
```
alismith-dev, line 541: Forgot to add the dependencies array! 🙂 

```suggestion
  }, []);
```

alismith-dev:
@LucyMac I took the liberty of rebasing (well, actually cherry-picking was easier) your branch onto master.

alismith-dev:
@LucyMac I'm going to go ahead and merge this now. We can follow up later with the suggested fixes.

PR #417, opened 2020-05-05
React week 3: "advanced useEffect" section

alismith-dev:
Replaces class-based lifecycle sections with "advanced `useEffect`" sections.

#### Unmounting & cleaning up effects

Returning an unsubscribe function from the effect.

I'm a little unsure if this should be included or moved to stretch goals. I think it might be worth trying, but keeping an eye on the time. If we have time, I think it is worth it.

#### Updating data when props change

Using the dependencies array to re-run effects when props change.

I decided to frame this by showing a "broken" component first (i.e. one that doesn't update effects in response to props changes). Then "playing computer" to explore why it isn't working, before showing the solution. I think this explains why the solution works nicely. As noted in the mentor notes, although each steps are included in the syllabus, I was intending on just using the CodeSandboxes to actually do the walk-through. The steps are included just for revision purposes.

I also included a small UX optimisation to `useEffect`, to show a loading state when the effect is re-run after props change. If short on time, I think this could probably be skipped.

--- 

For simplicity of review, here's [the rendered version of the week 3 lesson](https://github.com/CodeYourFuture/syllabus/blob/e5a8ec735e3424c0c06daf89834eba0c444df832/react/week-3/lesson.md).

@LucyMac @anthonytranDev @frosas


alismith-dev:
As discussed, I have removed the unmounting/cleaning up effects sections. Merging.

PR #418, opened 2020-05-06
Add instructions on stopping create-react-app and installing stop-runaway-react-effects

alismith-dev:
Does exactly what it says on the tin.

### Stopping create-react-app

This is needed because CRA tends to consume a lot of resources on old laptops, so stopping when not needed is recommended. This is worse if they accidentally run multiple instances (e.g. in a closed background tab).

### Installing stop-runaway-react-effects

To help prevent a mistake with `useEffect` which causes an infinite loop, install [`stop-runaway-react-effects`](https://github.com/kentcdodds/stop-runaway-react-effects) which prevents this problem.

### DIFF
--- a/react/week-1/lesson.md
+++ b/react/week-1/lesson.md
@@ -139,51 +139,8 @@ As you can see, this is much easier to read than both the straight `React.create
 
 ## Let's Create A React App
 
-The Facebook team behind React have created a tool to help you create and set up React projects. It is called [Create React App](https://create-react-app.dev/). It sets up files like we saw in the previous example, so that you don't have to.
-
-> **Exercise C**: Install & set up a Create React App by following the steps below
-
-```
-npx create-react-app pokedex
-cd pokedex
-```
-
-Open the `pokedex` directory in your editor. Notice that create-react-app has created a bunch of folders for you. We will be working with the `pokedex` application for the rest of the module.
-
-### Starting the app
-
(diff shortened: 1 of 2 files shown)


react/week-1/lesson.md
```diff
@@ -148,10 +148,32 @@ npm install -g create-react-app
...
+
+```
 npm start
 ```
 
-The last command should open a web browser for you. This shows the application that we are going to be working on. Open the `pokedex` directory in your editor. Notice that create-react-app has created a bunch of folders for you. We will be working with the `pokedex` application for the rest of the module.
+This does two things:
+
+1. Run a program on your computer that *watches* your files and updates your application when you make changes. It also runs some checks for common bugs or problems in your code. An error message will be shown in your browser if it detects that there is a problem
+2. Opens a web browser with a link to your React app so that you can test it
```
zubhav: Can we change `so that you can test it` to `so that you can develop and test any changes made` or something along that line?

LucyMac (approved):

react/week-1/lesson.md
alismith-dev: Done

alismith-dev:
Thanks both, as discussed on the call, I've moved the content to [the wiki](https://docs.codeyourfuture.io/students/guides/creating-a-react-app) so this PR now just links the syllabus to the wiki. Merging.

PR #421, opened 2020-05-08
JS Core 1 

### LINKED ISSUE #420: JavaScript Core 1 revisions
Rework JS1 so it is suitable for students with no previous coding experience.

Following these principles -
1. Don't use anything in examples or exercises which hasn't been specifically taught yet
2. Build up students' confidence & comprehension by starting each section with an easy exercise and building slowly
3. Use simple English when possible

r-darby:
[Issue #420](https://github.com/CodeYourFuture/syllabus/issues/420) 

Week 2
In Loops section separate "if" and "while" into 2 sections
Arrays 2nd exercise - simplify exercise so we don't remove an element
Arrays 4th exercise - simplify
Extra exercises - add a first exercise which is simpler

Week 3
Simplify exercises for Callback functions, Chaining, forEach

### DIFF
--- a/js-core-1/week-2/lesson.md
+++ b/js-core-1/week-2/lesson.md
@@ -400,7 +400,7 @@ Write a function, similar to the last exercise, that:
 
 - Takes one number `n` as a parameter
 - Adds all numbers from `0` to `n`. For example, if the input is `3`, the output should be `0 + 1 + 2 + 3`
-- You should use an if loop
+- You should use an for loop
 
 ## Arrays
 
@@ -457,14 +457,15 @@ In the node REPL, enter the following array:
 > const fruits = ['banana', 'apple', 'strawberry', 'kiwi', 'fig', 'orange'];
 ```
 
-Which index do you need to return:
+1. Now, using the correct indexes, get the following values from the array:
 
 - strawberry
 - kiwi
 - orange
 - banana
 
-Replace 'apple' with 'raspberry' and 'fig' with 'pineapple'
+
+2. Replace 'apple' with 'raspberry' and 'fig' with 'pineapple'
 
 ### Exercise (5 mins)
 
@@ -479,7 +480,7 @@ function secondMatchesAmy(array) {
 }
 ```
 
-### Using if loops with arrays
+### Using for loops with arrays
 
 We can use the power of loops to run some code for each element in our array.
 
--- a/js-core-1/week-2/lesson.md
+++ b/js-core-1/week-2/lesson.md
@@ -357,6 +357,8 @@ In pairs, write a function that checks a username is of an acceptable format for
 
 When we're writing programs, we often find that we want to repeat a bit of code over and over, or repeat it but change something about it each time. To save ourselves from writing all that code, we can use a **loop**. JavaScript has two kinds of loops, a `while` loop and a `for` loop.
 
+### while loop
+
 A `while` loop is a way to repeat code until some condition is evaluated to `false`. For example:
 
 ```js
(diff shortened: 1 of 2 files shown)


js-core-1/week-2/lesson.md
```diff
@@ -472,7 +479,27 @@ function secondMatchesAmy(array) {
 }
 ```
 
-### Exercise (20 mins)
+### Using if loops with arrays
```
illicitonion:
```suggestion
### Using for loops with arrays
```

js-core-1/week-3/lesson.md
```diff
@@ -364,19 +363,11 @@ names.forEach(function (name, index) {
...
-}
+### Exercise (10 minutes)
 
-names.map(formatName).forEach(function (name, index) {
-  console.log(index + ": " + name);
-});
-```
+- Create a function that takes a `birthYear`, and returns the age of someone
+- With an array of ages, create another array filled with their birth years `[ 1964, 2008, 1999, 2005, 1978, 1985, 1919 ]`
+- `console.log` the birth years array
```
illicitonion: I remember this exercise being kind of confusing in class. I _think_ this line is the wrong way around - it's meant to be saying:

```suggestion
- With an array of birth years `[ 1964, 2008, 1999, 2005, 1978, 1985, 1919 ]`, create another array filled with their ages.
- `console.log` the ages array
```

illicitonion (approved):
This is looking great, thanks for cleaning it up!

js-core-1/week-2/lesson.md
```diff
@@ -381,13 +394,13 @@ for (let i = 0; i < 20; i++) {
...
 
-Write two functions that:
+Write a function, similar to the last exercise, that:
 
-- Take one number `n` as a parameter
-- Add all numbers from `0` to `n`. For example, if the input is `3`, the output should be `0 + 1 + 2 + 3`
-- The first version of the function should use `while` and the second one `for`.
+- Takes one number `n` as a parameter
+- Adds all numbers from `0` to `n`. For example, if the input is `3`, the output should be `0 + 1 + 2 + 3`
+- You should use an if loop
```
djgrant: I think you mean `for` here!

js-core-1/week-2/lesson.md
```diff
@@ -444,20 +457,14 @@ In the node REPL, enter the following array:
 > let fruits = ['banana', 'apple', 'strawberry', 'kiwi', 'fig', 'orange'];
```
djgrant: We should change this to `const` in case it implicitly teaches that you need `let` to mutate an array.

djgrant: I know different people use `const` and `let` in different ways but it's normally either `const` as default or `let` as default. Everywhere else we have `const` as default.

js-core-1/week-2/lesson.md
```diff
@@ -444,20 +457,14 @@ In the node REPL, enter the following array:
...
 ```
 
-1. Which index do you need to return:
+Which index do you need to return:
```
djgrant: This question is a bit ambiguous. I read it as though I'm being asked to return an index. 

How about:

> Now, using the correct indexes, get the following values from the array:

js-core-1/week-2/lesson.md
```diff
@@ -472,7 +479,27 @@ function secondMatchesAmy(array) {
...
+### Using if loops with arrays
+
+We can use the power of loops to run some code for each element in our array.
+
+When we do this say we **iterate** over an array.
+
+```js
+let daysOfWeek = ["Monday", "Tuesday", "Wednesday", "Thursday", "Friday", "Saturday", "Sunday"];
+
+for(let i = 0; i < daysOfWeek.length; i++){
```
djgrant, line 492: I think this is a good opportunity to highlight the difference between `const` and `let`. If everything uses `const` by default we can simply ask, what happens if you can `let i = 0` to `const i = 0`? They will get an error saying that you can't reassign `const` and there's your answer!

js-core-1/week-3/lesson.md
djgrant: Could we make this even clearer:

> You are given an array with year that 7 people were born, `[ 1964, 2008, 1999, 2005, 1978, 1985, 1919 ]`. Take this array and create another array containing their ages.

djgrant (changes requested):

djgrant:
Seems that from this module onwards the syllabus changes to using `let` as default. I think this way round will probably create extra confusion but perhaps to late to be changing it.

js-core-1/week-2/lesson.md
r-darby: Indeed, that could've been awkward! Many thanks

PR #426, opened 2020-05-09
Add Handling Events section to React Week 2

alismith-dev:
Adds a section about handling events to week 2 of React module. Previously I felt like we glossed over this a bit, so I think it's worth discussing for a bit.

[Rendered version](https://github.com/CodeYourFuture/syllabus/blob/react-handling-events/react/week-2/lesson.md).

While I was here, I decided to remove most of the class components content from week 2 as well.

### DIFF
--- a/react/week-2/lesson.md
+++ b/react/week-2/lesson.md
@@ -50,149 +50,104 @@ const HelloMentor = () => (
 );
 ```
 
-## Class Components
+## Handling events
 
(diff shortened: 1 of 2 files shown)


alismith-dev:
I'm fairly happy with where this has got to, so I'm going to merge it now. If there's a strong objection to this content, I'm open to feedback on moving it/splitting it up/making it optional, so please open a discussion on Slack 🙂 

Merging.

PR #429, opened 2020-05-10
Update React week 2 section to be clearer that it is about re-rendering

alismith-dev:
Updates the existing Reacting to Changes section to rename to Re-Rendering Components. This also removes some unnecessary details from the example code about how a forced re-render was happening. The section attempts to discuss re-rendering separate from state, but the primary focus is now on why re-rendering is an important concept.

[Rendered version](https://github.com/CodeYourFuture/syllabus/blob/react-re-rendering/react/week-2/lesson.md).


zubhav (approved):
👍

PR #430, opened 2020-05-10
Convert React Forms section to use Hooks

alismith-dev:
Does exactly what it says on the tin.

Most of these changes I think are uncontroversial. However I wasn't sure exactly how to structure the state for the form with multiple fields. I ended up going for a single object:

```js
const [userData, setUserData] = useState({
    username: "",
    email: "",
    password: ""
  });
```

But this means we have to use the object spread operator to update the state:

```js
const updatedUserData = {
  ...userData,
  [event.target.name]: event.target.value
};

setUserData(updatedUserData);
```

I have included a section explaining the spread operator, but I have concerns that it might be a big leap for some students.

An alternative approach would be to abandon the dynamic object key idea and just use an `if`/`switch`:

```js
if (event.target.name === "username") {
  setUsername(event.target.value)
} else if (event.target.name === "email") {
  setEmail(event.target.value)
} else if (event.target.name === "password") {
  setPassword(event.target.value)
}
```

What do you think?

[Rendered version](https://github.com/CodeYourFuture/syllabus/blob/react-forms-hooks/react/week-3/lesson.md).

### DIFF
--- a/react/week-3/lesson.md
+++ b/react/week-3/lesson.md
@@ -483,21 +483,24 @@ We are combining all of the concepts above to make a new object, that has all th
 
 | **Exercise D** |
 | :--- |
-| 1. Open the `pokedex` React application again and open the `src/CaughtPokemon.js` file. In this exercise, instead of recording the number of caught Pokemons, we are going to record the names of each Pokemon you caught. |
-| 2. Make sure the `CaughtPokemon` component is written as a class component. |
-| 3. Add an `<input>` in the `render` method before the `button` (hint: `<input type="text" />`). |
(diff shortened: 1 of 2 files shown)


zubhav:
I really think seperate `useState`s are the best approach for this. It will give students practice in managing multiple states, which is probably more common than managing a single object.

alismith-dev:
I'm going to merge this as it is now (i.e. with the object spread) if there's strong objections we can change it later.

If we find that they really struggle with it, I think we can change tack and teach multiple states. Or possibly just mutating the object locally before setting that on state.

PR #432, opened 2020-05-11
Add Learning Objectives to JS2-Week 1

ChrisOwen101:
@Abdoulrazack95 wrote these learning objectives for JS2-1

```
From the end of today I will be able to answer the following …
1- What is an Object in Javascript and how to use it ? 
2- How to show the value of an Object’s property and change the value of that property ?
3- How Objects methods work ?
4- What are Arrays of Objects ?
5-What are Object built-in methods and how to use them ?
6- How to iterate through an Object ? 
```

I've re-written them slightly since it is best if the LOs are defined in what the student can do rather than what the student can know.

### DIFF
--- a/js-core-2/week-1/lesson.md
+++ b/js-core-2/week-1/lesson.md
@@ -22,6 +22,21 @@ Read the Mentors Notes [here](./mentors.md)
 
 ---
 
+## Learning Objectives
+
+By the end of this lesson students should be able to:
+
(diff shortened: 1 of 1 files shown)


AlinChindea (approved):
🚀

lannem (approved):
:)

Abdoulrazack95 (approved):

PR #436, opened 2020-05-12
[DATABASES] Add Learning Objectives for SQL

ChrisOwen101:
# What does this change?

Module: Databases SQL
Week(s): All three

## Description

This PR adds learning objectives to the Database SQL module

@llh1 @agiamas

### DIFF
--- a/db/week-1/lesson.md
+++ b/db/week-1/lesson.md
@@ -29,6 +29,22 @@ Read the Mentors Notes [here](./mentors.md)
 
 ---
 
+## Learning Objectives
+
+By the end of this lesson students should be able to
+
+- Explain why databases are needed in Web Development
+- Define what SQL is and what it is used for
+- Describe what `table`, `rows` and `columns` are used for in a relational database
+- Install and create a database using PostgreSQL
+- Create a table in a database using PostgreSQL
+- Insert data into a table using PostgreSQL
+- Retrieve data from a table using PostgreSQL
+- Retrieve data from a table using conditionals in PostgreSQL
+- List the different kinds of data that can be held in a PostgreSQL database
+
+---
+
 ## Introduction to databases
 
 A database is a structured set of data held in a computer. It provides ways to store, retrieve and organize information.
--- a/db/week-3/lesson.md
+++ b/db/week-3/lesson.md
@@ -21,6 +21,18 @@ Read the Mentors Notes [here](./mentors.md)
 
 ---
 
+## Learning Objectives
+
+By the end of this lesson students should be able to
+
+- Create an API endpoint in a NodeJS applicant to add data to a database
+- Write code that effectively validates data before adding it to a database
+- Create an API endpoint that can retrieve data from a database
+- Create an API endpoint that can update data in a database
+- Create an API endpoint that can delete data in a database
+
+---
+
 ## Recap integration of cyf_hotels DB with NodeJS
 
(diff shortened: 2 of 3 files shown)


db/week-3/lesson.md
```diff
@@ -21,6 +21,18 @@ Read the Mentors Notes [here](./mentors.md)
 
 ---
 
+## Learning Objectives
+
+By the end of this lesson students should be able to
+
+- Create an API endpoint that can add data to a database
+- Write code that effectively validates data before adding it to a database
```
illicitonion, line 29:
```suggestion
- Write code that effectively validates data before adding it to a database
- Be aware of the dangers of SQL injection attacks, and how to avoid them
```

db/week-1/lesson.md
```diff
@@ -29,6 +29,22 @@ Read the Mentors Notes [here](./mentors.md)
...
+By the end of this lesson students should be able to
+
+- Explain why databases are needed in Web Development
+- Define what SQL is and what it is used for
+- Describe what `table`, `rows` and `columns` are used for in a relational database
+- Install and create a database using PostgreSQL
+- Create a table in a database using PostgreSQL
+- Insert data into a table using PostgreSQL
+- Retrieve data from a table using PostgreSQL
+- Retrieve data from a table using conditionals in PostgreSQL
```
illicitonion, line 43: Do we also want to cover ordering and/or aggregates and/or group-by? e.g. `SELECT * FROM t ORDER BY c LIMIT 1` and/or `SELECT MAX(c) FROM t` and/or `SELECT name, COUNT(*) FROM t GROUP BY name`

illicitonion (approved):
Looks good :) Added a couple of things we may want to think about covering

db/week-1/lesson.md
llh1: Some of these may come in the week 2 class @illicitonion , it's not part of the first class.

llh1 (approved):
LGTM 👍

db/week-1/lesson.md
illicitonion: Yeah, I'm not sure _where_ these should be, but they're currently not listed in any of the three :) Week 2 probably makes sense

KeithBremer: I'm not sure why we're teaching CREATE TABLE in week 1 - the students have no experience of using tables at this stage so understanding what's needed to create one is much harder.  We also then have to teach SQL datatypes, etc.  Wouldn't it be better to concentrate on queries initially then progress to insert, update & delete later (e.g. week 2) and create table only after we've covered JOIN and primary/foreign key (maybe week 3).  I'm not sure how many front-end developers would ever need to use create table anyway.

llh1: Hey :) At this stage, students have no experience whatsoever using SQL anyway. In Glasgow, in the past 2 years, we started with a paper exercise, by thinking about how to organise data for a problem so it was really nice to introduce the concepts of table, columns, rows, keys and types. Overall it worked really well, and it was one of the module where students had a lot of success to understand what's going on because everything is well structured in SQL. We are not teaching people to be frontend developer more than backend developer, at least in Glasgow, so i really think introducing  how things are stored first help a lot. If you start with insert, you still need to understand the concept of table, column and types, isn't it? So in my opinion, the most logical way is to actually start with creating tables :)

PR #437, opened 2020-05-12
[React] Lesson 1 tweaks

LucyMac:
# What does this change?

Module: React
Week(s): 1

## Description

- Reduces examples to focus on core parts
- Suggested change for installing CRA to follow recommended use of npx?
- tweaks copy a little
- fixes typos


react/week-1/lesson.md
```diff
@@ -110,43 +86,29 @@ JSX is a simple syntax *sugar* that looks like HTML, but is actually converted t
...
 > **Exercise C**: Install & set up a Create React App by following the steps below
 
-```
-npm install -g create-react-app
+1. Open the command line and navigate to where you usually save projects
+2. Run the code below:
 
-create-react-app pokedex
+```
+npx create-react-app pokedex
```
zubhav: Just a note for the use of `npx`. This is available for npm@5.2.0 or higher. Given that the current version of npm is 6+ now, I think this should be safe to assume. But, I thought I'd note this just incase someone has an issue. 👍

react/week-1/lesson.md
```diff
@@ -277,10 +241,8 @@ Now instead of hard-coding the greeting in the `Greeting` component, we are usin
 
 ```js
 const Mentor = () => {
-  const mentors = ['Ali', 'Kash', 'Davide', 'German', 'Gerald'];
-  return (
-    <span>{mentors.join(', ')}</span>
-  );
+  const mentors = ["Ali", "Sub", "Loic", "Anthony", "Lucy", "Mozart"];
```
zubhav, line 231: 👏

zubhav (approved):

react/week-1/lesson.md
LucyMac: Good point. I'll check what we know about people's setup 👍

alismith-dev: This has already been updated on master: https://syllabus.codeyourfuture.io/react/week-1/lesson.html#lets-create-a-react-app. For simplicity of pre-installing, it has been moved to the wiki: https://docs.codeyourfuture.io/students/guides/creating-a-react-app.

Re: versioning, we made sure the students installed a recent version of Node (v13.x) when they started JS1. It comes bundled with npm 5+

react/week-1/lesson.md
```diff
@@ -59,47 +59,23 @@ Let's take a look at an example. We are going to walk through how to render a `<
...
+divNode.innerText = "Hello World";
+
+var rootElement = document.querySelector("#root");
+rootElement.appendChild(divNode);
 ```
 
 Now let's convert to using React ([interactive version](https://jsbin.com/calaqitede/1/edit?html,output)):
 
-```html
-<!DOCTYPE html>
```
alismith-dev, line 82: The original intention was to show that React isn't "magic", as CRA hides away a lot of this stuff.

But on reflection I think I agree with this change. The HTML just adds cognitive load. 👍

alismith-dev (approved):
Thanks @LucyMac, especially for catching all my typos 😆

Re: merging onto which branch, I've left a comment about using CRA, which has already been updating on `master`.

react/week-1/lesson.md
LucyMac: The JSBin examples still have the complete version 👍

LucyMac (commented):
Removed CRA install instructions

PR #431, opened 2020-05-10
Syllabus Update - React Week 2 - useState

anthonytranDev:
Draft alterations have been made from previous reviews see https://github.com/CodeYourFuture/syllabus/pull/405

PR #448, opened 2020-05-19
Updates to useState content

alismith-dev:
# What does this change?

Module: React
Week(s): 2

## Description

Originally cherry-picked from https://github.com/CodeYourFuture/syllabus/pull/431, but based on `master` not `london`. I'm taking over ownership of this as Anthony doesn't have time. I've made some changes, but the core is still there.

I also removed most of the old classes content about state as it is no longer needed.

**Note**: there are no exercises currently. These will be done in a follow-up PR.

[Rendered version](https://github.com/CodeYourFuture/syllabus/blob/react-use-state-updates/react/week-2/lesson.md)

@anthonytranDev
@LucyMac


react/week-2/lesson.md
```diff
@@ -193,270 +181,297 @@ We don't need to worry about changing the DOM ourselves! This is what makes Reac
 
 ## State
 
-Let's take another look at the the counter example. What if you wanted to create multiple counters on the same page? How would you add another counter?
+*State* a general concept in software engineering. It is used when part of your app needs to "remember" something that changes when people interact with it.
```
zubhav:
```suggestion
*State* is a general concept in software engineering. It is used when a part of your app needs to "remember" something that changes when people interact with it.
```

react/week-2/lesson.md
```diff
@@ -193,270 +181,297 @@ We don't need to worry about changing the DOM ourselves! This is what makes Reac
...
-        <button>Click me!</button>
-      </div>
-    );
-  }
-}
+let useState = React.useState;
 ```
 
-This code has a bug! `this.state` is initialised as an empty object, and so `this.state.count` is undefined. We need to initialise it from props. We can do this in the class constructor ([interactive example](https://codesandbox.io/s/1oyxx4lzz7)):
+In fact we can just write `React.useState` in our component if we want! But to type a bit less code, we import it (using the curly braces) once and then can just use `useState`.
```
zubhav, line 229: Shall we show an example here of the curly braces method?

`import React, { useState } from 'react'`

react/week-2/lesson.md
```diff
@@ -193,270 +181,297 @@ We don't need to worry about changing the DOM ourselves! This is what makes Reac
...
+So far we've only seen an example with one state variable. But you can create multiple state variables if you want! Let's see an example ([interactive example](https://codesandbox.io/s/multiple-state-variables-piq5w?file=/src/FruitCounter.js)):
+
+```js
+function FruitCounter() {
+  const [fruit, setFruit] = useState("bananas");
   const [count, setCount] = useState(0);
 
+  function handleOrangesClick() {
+    setFruit("oranges");
+  }
```
zubhav, line 373: Nit-picking here- lets add a new line after this function to separate it from the next one

react/week-2/lesson.md
```diff
@@ -520,15 +535,23 @@ This is a very common pattern which will come in very useful!
 | 12. Still within the second `.then` callback function, update the `pokemon` state variable. <details><summary>Click here is you are stuck.</summary>:ppl at the State section again to see how to set state variables to new values.</details> |
 | 13. What happens in your browser? Do you understand why? If not, discuss it with another student. If you are both stuck, ask a Teaching Assistant. |
 
+### Container components
+
+In real world applications, the things we want to remember in state follow the *business logic* required by our users. So for example the number of caught Pokemon in the exercise  increases when you click on the button *Catch Pokemon*. Most of the time, business logic is about figuring out when and how to change state.
```
zubhav: Will the students understand the term 'business logic' or should we be more explicit here about what that is?

react/week-2/lesson.md
alismith-dev: I thought about suggesting that we only use `React.useState`/`React.useEffect` but my sense is that they'll need to know it anyway since lots of tutorials/production code use this import method.

react/week-2/lesson.md
alismith-dev: While in prod code I'd normally agree with you, this was intentional. I'm attempting to group the fruit handlers separate from the count handler.

react/week-2/lesson.md
alismith-dev: Yes good point. Probably worth a link to a definition.

react/week-2/lesson.md
```diff
@@ -193,270 +181,297 @@ We don't need to worry about changing the DOM ourselves! This is what makes Reac
...
-let count3 = 0;
-```
+1. We'll set the initial state (before any interaction as happened):
 
-What might be the problem here?
+  ```js
+  state = {
+    buttonPressed: false
+  }
+  ```
```
LucyMac: Could we maybe use a more concrete scenario such as:

```js
    userIsLoggedIn: false;
```

Alternatively we can always provide real-world examples verbally in class of when we might use state. Just trying to make it a little less abstract.

By the by, it makes me a little nervous when the syllabus contains bits of code that do not reflect what students should actually be doing i.e. creating an object called `state`. I see how it's useful to take them on the journey to understand why state shouldn't be a variable though, and your next section does that. It's a tricky one.

react/week-2/lesson.md
```diff
@@ -193,270 +181,297 @@ We don't need to worry about changing the DOM ourselves! This is what makes Reac
...
-    </div>
-  );
-};
+const [count, setCount] = useState(0);
 ```
 
-This still isn't a particular useful application, because we can only still only count to 1! We need to change our `Counter` component so that it reads the previous state, then adds 1 onto that. We can do this by passing a callback function to `this.setState` ([interactive example](https://codesandbox.io/s/qxz27q9y4)):
+To understand this bit of code, we first have to understand *destructuring*.
+
+### Destructuring Arrays
```
LucyMac: Could we maybe move this to somewhere else, and link to it here? 🤔 can still be taught in class.

react/week-2/lesson.md
```diff
@@ -193,270 +181,297 @@ We don't need to worry about changing the DOM ourselves! This is what makes Reac
...
 
-### Container components
+```js
+function Counter() {
+  const [count, setCount] = useState(0);
 
-In real world applications, the things we want to remember in state follow the *business logic* required by our users. So for example the number of caught Pokemon in the exercise  increases when you click on the button *Catch Pokemon*. Most of the time, business logic is about figuring out when and how to change state.
+  function incrementCount() {
+    const newState = count + 1;
+    setCount(newState);
```
LucyMac: Might it be confusing for students to see this `newState` object here? It makes it look like we might always need to create this `newState` object / confuse the `state` concept. 
Can we just `setCount(count + 1)` ?

We could always add a comment
```js
  function incrementCount() {
    // whatever gets passed to the setCount function will become the new state
    setCount(count + 1);
  }
```

react/week-2/lesson.md
```diff
@@ -193,270 +181,297 @@ We don't need to worry about changing the DOM ourselves! This is what makes Reac
...
+We have talked about how a component "remembers" state. In fact, each component remembers **separate** state from other components. This means we can have multiple different Counters, each with a different state ([interactive example](https://codesandbox.io/s/multiple-counters-xm1x4?file=/src/App.js)):
 
 ```js
-import React, { useState } from 'react';
+function App() {
+  return (
+    <div>
+      <Counter />
+      <Counter />
+      <Counter />
```
LucyMac, line 356: I'm not sure this example really illustrates the idea of each counter having its own state.
Maybe we could add comments next to each one?

```js
function App() {
  return (
    <div>
      <Counter />   // this counter's 'count' state could be 8
      <Counter />   // while this counter's 'count' state could be 0
      <Counter />   // and this one could be 42
    </div>
  );
}
```

react/week-2/lesson.md
```diff
@@ -193,270 +181,297 @@ We don't need to worry about changing the DOM ourselves! This is what makes Reac
...
+```js
+function FruitCounter() {
+  const [fruit, setFruit] = useState("bananas");
   const [count, setCount] = useState(0);
 
+  function handleOrangesClick() {
+    setFruit("oranges");
+  }
+  function handleBananasClick() {
+    setFruit("apples");
```
LucyMac: Do we mean to say 'apples' here and not 'bananas'?

LucyMac (commented):
Looks good @40thieves, thanks!
Couple of comments below, plus I wondered if we could show what passing functions into components looks like _with arguments_, and in the same vein introduce the syntax `onClick={() => handleClick()}` ? Though could be an additional bit at the end.

react/week-2/lesson.md
alismith-dev: Already fixed :)

react/week-2/lesson.md
alismith-dev: I was just going to rely on opening the CodeSandbox and demo-ing it (I do this with basically all code snippets).

Comments in JSX are kind of a pain (you'd have to use `{/* stuff */}`). So I'm not sure it's worth it?

react/week-2/lesson.md
alismith-dev: There's perhaps a discussion to be had about moving this earlier in the syllabus. But I think if we're going to include it, it probably should be in the syllabus (it gets lost otherwise).

alismith-dev: Actually on second thoughts, I'd be happy to link to it. I'm just not sure where we'd move it to.

react/week-2/lesson.md
alismith-dev: > Could we maybe use a more concrete scenario 

Yep good point. I like the `userIsLoggedIn` example.

> By the by, it makes me a little nervous when the syllabus contains bits of code that do not reflect what students should actually be doing

I think maybe a diagram would be useful here?

> I see how it's useful to take them on the journey to understand why state shouldn't be a variable though, and your next section does that.

I ended up removing that section as it just confuses people.

react/week-2/lesson.md
alismith-dev: I made this change quite late. I originally did it to emphasise that whatever you pass to `setCount` will be stored (i.e. it doesn't have to be a calculation). I do agree the naming is confusing though.

I'll revert that and go back `setCount(count + 1)`. We can always pull that statement out in the lesson if there's questions about it.

react/week-2/lesson.md
alismith-dev: Replaced with a diagram in https://github.com/CodeYourFuture/syllabus/pull/448/commits/85645dbca58df5b52ef9c32eeb0a00355d5ab378

react/week-2/lesson.md
alismith-dev: Done in https://github.com/CodeYourFuture/syllabus/pull/448/commits/a247ddff9371df276917aa36d9760df26fa6c82f

react/week-2/lesson.md
LucyMac: Ok

react/week-2/lesson.md
LucyMac: sweet

react/week-2/lesson.md
LucyMac: There's a mention of array destructuring at the bottom of the lesson. Maybe we move it down there and add a link to anchor back to the relevant part of the lesson?

react/week-2/lesson.md
```diff
@@ -187,276 +188,289 @@ function Counter(props) {
...
+We won't look at how this works at the moment, but behind the scenes there is some code that will listen for clicks on the button and force React to update. That means when you click the button, the function component is called again (or *re-rendered*).
+
+Now `props.likeCount` is **1**. React now **updates** the DOM to make sure it shows the correct number. Every time we click the button, the function component is called and React updates the DOM for us.
 
 We don't need to worry about changing the DOM ourselves! This is what makes React so powerful. Even better, React will figure out exactly the right bits of the DOM that need to be changed, a concept called the ["virtual DOM"](https://reactjs.org/docs/faq-internals.html). This makes it extremely efficient and fast.
 
 ## State
 
-Let's take another look at the the counter example. What if you wanted to create multiple counters on the same page? How would you add another counter?
+*State* is a general concept in software engineering. It is used when a part of your app needs to "remember" something that changes when people interact with it.
```
mozartdiniz, line 199: Hey, @40thieves I liked the way you introduce state as a general component concept. I would like to suggest a slightly different way to introduce some of the concepts we are working here. Let me know what do you think.

## State

One of the many strategies that React uses to decide when is the right time to reflect any change on the DOM is to watches a specific set of variables. Those variables are called state. It means that not everything I change inside a component will result in a DOM change? Yeap, that is correct. 

```js
function Counter(props) {
  let counter = 0;

  setTimeout(() => {
     counter = counter += 2;
  }, 1000); 

  return <div>{ counter }</div>;
}
```

What should we expect to see when we run this code and wait for one second? Spoiler! Nothing will happens. Why? The reason why is we are setting a regular variable inside this component, React don't care about regular variables. We could change it to a string, array, object, none of those changes would make this component change from 0 to two in after one second.

So, how do we make React watch to our changes? We have to use a state instead.

**But what is a state?** 

*State* is a general concept in software engineering. It is used when a part of your app needs to "remember" something that changes when people interact with it.

State diagram

This is a simple example, but if we had lots of bits of state, then we can make very complex apps.

React provides a function to help use to use states. it is conveniently called useState. We have to import it from React packages.

```js
import React, { useState } from 'react';

console.log(useState)
```

If we look at the console, `useState` is just a function. It lives inside the React code that you installed when you created the app.

To reference the `useState` function in our component, we need to import it from the React code. The curly braces around `useState` are a bit like writing:

```js
let useState = React.useState;
```

In fact we can just write `React.useState` in our component if we want! But to type a bit less code, we import it (using the curly braces) once and then can just use `useState`.

### Using `useState`

Let's refactor the function above to use state rather than `let`.

```js
function Counter(props) {
  const [count, setCount] = useState(0);

  setTimeout(() => {
     setCount(count + 2);
  }, 1000); 

  return <div>{ counter }</div>;
}
```

alismith-dev: Hi Mozart 🙂 Thanks for the suggestion!

Anthony had (and I adapted) a similar section about the difference between variables and state. The easiest way to see it is in this commit: https://github.com/CodeYourFuture/syllabus/pull/448/commits/ae96a11606d8013a20df38040d5be736f49a8d07

I decided to remove it since Neill pointed out that it may be confusing that the first code example in the state section doesn't actually work. (See the very long feedback thread in Slack). I am also somewhat concerned about the amount of material we have to get through, so decided to cut it.

I could however be persuaded, as we can now point to the cheatsheet for copy/paste code snippets. The syllabus here I think is intended to be mostly read in large chunks, not as a reference guide.

---

If you look at the old class-based state lesson (the easiest way is probably to look at the backed-up scotland syllabus: https://codeyourfuture.github.io/syllabus-scotland/react/week-20/lesson.html#reacting-to-changes) you'll see that I had a similar idea previously - refactoring a broken app to use state.

The feedback generally was that this problematic because:

1. Similar to above, it teaches the "wrong way" first
2. Students generally found my explanation of forced re-rendering confusing

I tried to capture the essence of this in the new Re-Rendering Components section: it demos a similar app, although it does actually work (by re-calling `ReactDOM.render`).

react/week-2/lesson.md
alismith-dev: Ok yeah. I think as long as we add it to the cheatsheet (for easier reference) that'll be fine.

alismith-dev: Done in https://github.com/CodeYourFuture/syllabus/pull/448/commits/bb354ca90551b8a6b2cb3a70bb5e3d90b54d4166

alismith-dev:
Although there's a few outstanding comments/ideas, I'm going to go ahead and merge this now. As discussed a while back, it's easier to make another PR than to debate in comments here.

PR #450, opened 2020-05-25
React week 2 in-class exercises

LucyMac:
# What does this change?

Module: React
Week(s): 2

## Description

- I took the liberty of swapping out IDs for classes to avoid any confusion and limit CSS bad practices (the IDs weren’t being used for anything other than giving a friendly name to buttons)
- I updated the code for the Martian photo example, and created a sandbox for it. @40thieves you might want to copy that to live in the same domain as the other sandboxes?
https://codesandbox.io/s/the-useeffect-hook-jtz5u?file=/src/MartianPhotoFetcher.js
- I changed the `pokemon` state variable name to be called `bestPokemon` instead, so we avoid the pattern `pokemon={pokemon}` which I find causes confusion as to which is which.
Speaking of which, I'm thinking we could do without the 'hint' in the final exercise which spells out how to pass the pokemon state down as a pokemon prop. I haven't made this change, it's just a suggestion, but I feel like it spoon-feeds it all a bit. 
- The final exercise with the pokemon data fetch feels unfinished. Should we add an instruction to pass down the pokemon from the API response and change the image + also the abilities? We can leave it quite vague and see what they come up with, but I feel we should explicitly say what bits of data they should be using at least.
- Removed a reference to class components down the bottom.


react/week-2/lesson.md
```diff
@@ -169,7 +169,7 @@ Let's look at how a component is re-rendered ([interactive version](https://code
 function Counter(props) {
   console.log(`Rendering. props.count is ${props.likeCount}`);
 
-  return <button id="like-button">Likes: {props.likeCount}</button>;
+  return <button class="like-button">Likes: {props.likeCount}</button>;
```
nbogie: className here

react/week-2/lesson.md
```diff
@@ -223,7 +223,7 @@ function Counter() {
   return (
     <div>
       Count is: {count}
-      <button id="click-me" onClick={incrementCount}>
+      <button class="click-me" onClick={incrementCount}>
```
nbogie: className here

react/week-2/lesson.md
```diff
@@ -493,26 +495,37 @@ useEffect(() => {
...
   }, []);
 
   return (
     <div>
-      // TODO: update this example
+      {marsPhotos.photos &&
+        marsPhotos.photos.map((photo, index) => {
+          return (
+            <img
+              key={`mars-photo-${index}`}
```
nbogie: just curious -  why not just {index for the key}?  that'd be just as unique in context and react doesn't care what it is.

react/week-2/lesson.md
```diff
@@ -522,26 +535,26 @@ This is a very common pattern which will come in very useful!
 | **Exercise** |
 | :--- |
 | 1. Open the `pokedex` React application again and open the `src/BestPokemon.js` file. |
-| 2. Create a new component called `BestPokemonFetcher`. |
+| 2. Add a new component inside `src/BestPokemon.js` called `BestPokemonFetcher`. |
 | 3. Change the `export default` to export `BestPokemonFetcher` instead of `BestPokemon`. We **don't** need to make any changes to the `BestPokemon` component. |
```
nbogie, line 514: Is it weird to have the *default* export (BestPokemonFetcher) not match the file name (BestPokemon)?  Newbie asking.  ( I see that pattern existed prior to this PR)

nbogie (commented):
Very minor stuff - class/className, and a question about key and default export-ing BestPokemonFetcher from BestPokemon.

react/week-2/lesson.md
LucyMac: omg I can't believe I made this mistake!

react/week-2/lesson.md
LucyMac: True it would be unique in this context. I thought I'd slip in a more realistic example of what often needs to be done to get a unique key in a larger app, where there might be more than one `map` on a page and therefore some duplicate indexes. 
Happy either way here, can revert to `key={index}` if we think it's clearer.

react/week-2/lesson.md
LucyMac: Probably, yes. We could either change the name of the file, or better still, create a new file entirely for the BestPokemonFetcher.

Or, moving it out into a separate component could be an exercise at the end after the 'Container component' section, to illustrate that `BestPokemonFetcher` is a container, separate from `BestPokemon` the presentational component

react/week-2/lesson.md
```diff
@@ -182,7 +182,7 @@ Let's look at how a component is re-rendered ([interactive version](https://code
 function Counter(props) {
   console.log(`Rendering. props.count is ${props.likeCount}`);
 
-  return <button id="like-button">Likes: {props.likeCount}</button>;
+  return <button className="like-button">Likes: {props.likeCount}</button>;
```
alismith-dev: Thanks for the suggestion here. In this case, I think an `id` is more appropriate. If you look at the [linked CodeSandbox](https://codesandbox.io/s/force-component-re-rendering-llow115pll?file=/src/index.js), you'll see that I'm using this `id` to attach an event listener to.

Generally I use `id`s for attaching JS events and `class`es for styling. Hence in this case, an `id`.

react/week-2/lesson.md
```diff
@@ -120,7 +120,7 @@ A full list of special event handler props is available [here](https://reactjs.o
 
 Just like with `addEventListener` we pass the function reference to `onClick` instead of calling the function. Think of it like this: we give the function to React, so that React can call our function when the element is clicked.
 
-| **Exercise** |
+| **Exercise A** |
```
alismith-dev, line 123: Thanks, I meant to go back through and do this 👍

react/week-2/lesson.md
```diff
@@ -492,55 +502,66 @@ useEffect(() => {
...
+    )
       .then(res => res.json())
       .then(data => setMarsPhotos(data));
-    });
   }, []);
 
   return (
     <div>
-      // TODO: update this example
+      {marsPhotos.photos &&
```
alismith-dev: I realise this is a very common & popular pattern in React, and we should have covered it in week 1, however I do think that this is simpler and easier to read:

```js
if (marsPhotos.photos) {
  return (
    <div>
      marsPhotos.photos.map(...)
    </div>
  )
} else {
  return <div />
}
```

It also points out that we're rendering a useless `<div>` if no photos are loaded.

react/week-2/lesson.md
```diff
@@ -492,55 +502,66 @@ useEffect(() => {
...
-| 12. Still within the second `.then` callback function, update the `pokemon` state variable. <details><summary>Click here is you are stuck.</summary>:ppl at the State section again to see how to set state variables to new values.</details> |
+| 12. Still within the second `.then` callback function, update the `bestPokemon` state variable. <details><summary>Click here is you are stuck.</summary>Refer to the State section again to see how to set state variables to new values.</details> |
 | 13. What happens in your browser? Do you understand why? If not, discuss it with another student. If you are both stuck, ask a Teaching Assistant. |
 
 ### Container components
 
-In real world applications, the things we want to remember in state follow the [*business logic*](https://en.wikipedia.org/wiki/Business_logic) required by our users. So for example the number of caught Pokemon in the exercise  increases when you click on the button *Catch Pokemon*. Most of the time, business logic is about figuring out when and how to change state.
```
alismith-dev, line 491: Gah stupid MBP keyboard inserting double spaces!!

react/week-2/lesson.md
```diff
@@ -443,6 +443,14 @@ function FruitCounter() {
 }
 ```
 
+
+| **Exercise C** |
+| :--- |
+| 1. Open the `pokedex` React application and open the `CaughtPokemon.js` file. |
+| 2. Create a new state variable called `totalCaught` and initialise it to `0` |
+| 3. Add a button to the component with an `onClick` handler which increments the `totalCaught` count. Refer to the examples above to help you. |
```
alismith-dev: I wonder if we could add some prediction here? Something simple like "before you click the button discuss with your classmates what will happen".

react/week-2/lesson.md
alismith-dev: Yes it is explicitly going against the "best practice" that we've asked them to follow here.

I think I was trying to avoid spending time making sure students had correctly linked up separate component files. Also note that this file is further extended in week 3. In hindsight I think having everything in 1 file is a bad idea.

I can make a change which change the instructions if you'd like.

react/week-2/lesson.md
```diff
@@ -492,55 +502,66 @@ useEffect(() => {
...
 | 3. Change the `export default` to export `BestPokemonFetcher` instead of `BestPokemon`. We **don't** need to make any changes to the `BestPokemon` component. |
-| 4. In the new `BestPokemonFetcher` component, create a new state variable called `pokemon` and initialise it to `null`. <details><summary>Click here if you are stuck.</summary>Look at the State section to see how to create state variables.</details> |
-| 5. If there is no `pokemon` state (hint: `if (!pokemon) {}`), then return `null` so that the component renders nothing. |
-| 6. If there is some `pokemon` state (`else {}`), then render the `BestPokemon` component and pass the `pokemon` state variable as the `pokemon` prop (hint: `<BestPokemon pokemon={pokemon} />`). |
+| 4. In the new `BestPokemonFetcher` component, create a new state variable called `bestPokemon` and initialise it to `null`. <details><summary>Click here if you are stuck.</summary>Look at the State section to see how to create state variables.</details> |
+| 5. If there is no `bestPokemon` state (hint: `if (!bestPokemon) {}`), then return `null` so that the component renders nothing. |
```
alismith-dev, line 516: If stick with the `data && data.map()` pattern I commented on above then we should make this match.

react/week-2/lesson.md
```diff
@@ -492,55 +502,66 @@ useEffect(() => {
...
-| 2. Create a new component called `BestPokemonFetcher`. |
+| 2. Add a new component inside `src/BestPokemon.js` called `BestPokemonFetcher`. |
 | 3. Change the `export default` to export `BestPokemonFetcher` instead of `BestPokemon`. We **don't** need to make any changes to the `BestPokemon` component. |
-| 4. In the new `BestPokemonFetcher` component, create a new state variable called `pokemon` and initialise it to `null`. <details><summary>Click here if you are stuck.</summary>Look at the State section to see how to create state variables.</details> |
-| 5. If there is no `pokemon` state (hint: `if (!pokemon) {}`), then return `null` so that the component renders nothing. |
-| 6. If there is some `pokemon` state (`else {}`), then render the `BestPokemon` component and pass the `pokemon` state variable as the `pokemon` prop (hint: `<BestPokemon pokemon={pokemon} />`). |
+| 4. In the new `BestPokemonFetcher` component, create a new state variable called `bestPokemon` and initialise it to `null`. <details><summary>Click here if you are stuck.</summary>Look at the State section to see how to create state variables.</details> |
```
alismith-dev, line 515: Thanks, changing to `bestPokemon` is a good idea.

alismith-dev:
>  I'm thinking we could do without the 'hint' in the final exercise which spells out how to pass the pokemon state down as a pokemon prop. I haven't made this change, it's just a suggestion, but I feel like it spoon-feeds it all a bit.

This is always a tricky balance. My experience is that stronger students don't really read the hints so aren't bothered by them. But weaker students need more hand holding (even if they refer to the example code we've just gone through).

I'd propose we keep spoon-feeding and see how it goes? EDIT: another alternative could be to put more things in `<details>` blocks as this hides them by default: stronger students shouldn't need to reveal them.

If we do keep it, would you be against breaking up the state exercise a bit into smaller steps.

alismith-dev:
> you might want to copy that to live in the same domain as the other sandboxes?

I can fork it and move it myself but that would require changing the URL. I believe you should be able to do it yourself. If you go to [the dashboard](https://codesandbox.io/dashboard/sandboxes), then you should see you're part of the CYF "team".

You can then drag the sandbox from "My Sandboxes" into this folder:

Screen Shot 2020-05-25 at 17 31 43

This gives everyone on the CYF team edit permissions (and saves lots of URL updates).

react/week-2/lesson.md
alismith-dev: Admittedly I'm also using the `id` for styling, but I figured it was less distracting to do that than add both an `id` and a `class`.

alismith-dev:
> The final exercise with the pokemon data fetch feels unfinished. Should we add an instruction to pass down the pokemon from the API response and change the image + also the abilities? We can leave it quite vague and see what they come up with, but I feel we should explicitly say what bits of data they should be using at least.

Hmm yes, that is perhaps in conflict with my "break every step down" philosophy. I would be in favour of adding a hint to say exactly what to pass (perhaps could "hide" it using the `<details>` trick?)

react/week-2/lesson.md
```diff
@@ -443,6 +443,14 @@ function FruitCounter() {
 }
 ```
 
+
+| **Exercise C** |
```
alismith-dev: Could we move this above "Where does State live"/"Setting multiple states"? Those 2 sections aren't required to complete the exercise and are really more of an FYI.

alismith-dev:
Looking through it, I'm not sure where we could add more, but it feels like only 4 exercises during the day is not enough?

As noted in this thread: https://github.com/CodeYourFuture/syllabus/pull/448#discussion_r429762378, I removed the content about destructuring arrays (and just linked to a blog post). We could do a quick exercise on it (maybe just in CodeSandbox/JSBin/where-ever)?

@nbogie I think suggested something about not mutating state? Idk though, after that section feels like the natural place for the first state exercise.

Perhaps "when to use props or state" too? Although to me this section also doesn't feel "important" enough to merit exercise time.

react/week-2/lesson.md
LucyMac: Might be worth returning something useful though such as a "nothing to display" message

react/week-2/lesson.md
LucyMac: I've changed it to `if () {}` and `else {}` in the other example so we can keep this here. But I do think we should show the `&&` and `? :` pattern during class.

react/week-2/lesson.md
LucyMac: I've split this exercise into exercises C and D

LucyMac:
In haste, it's late 💤 
I've split exercises and added some more, plus clarified some instructions.
I've also added a section on conditionals + a blurb on the `useState` setter function naming convention.

@40thieves / @nbogie / @mozartdiniz

react/week-2/lesson.md
LucyMac: FYI I've added a mini exercise after 'Container components' for students to have a think about where we could apply the container/presentational pattern in pokedex - and the 'right answer' would be that `BestPokemonFetcher` is a container while `BestPokemon` is presentational. It's at the very end of the day and if there's time we could get students to split those 2 components then, to implement the pattern?

react/week-2/lesson.md
LucyMac: I've added a section on conditional rendering incl different syntaxes.

react/week-2/lesson.md
LucyMac: Exercise split in half - half is before 'Where does State live', half is after.

react/week-2/lesson.md
alismith-dev: >  It's at the very end of the day and if there's time we could get students to split those 2 components then, to implement the pattern?

Sounds like a good plan 👍

react/week-2/lesson.md
alismith-dev: Thanks, looks good.

react/week-2/lesson.md
alismith-dev: I like the exercise, thanks (I was struggling to come up with a simple realistic that needed 2 states without replacing the counter entirely).

alismith-dev (approved):
Thanks for addressing all the comments @LucyMac! LGTM :shipit:

PR #456, opened 2020-05-28
Clarify not all function calls need to be assigned

illicitonion:
# What does this change?

Module: JS1
Week(s): Week 1

## Description

Clarify that function calls don't always need to be assigned to variables.

A student was reading this back to me as revision and got really confused about
trying to use a function _without_ assigning its returned value to a variable.

### DIFF
--- a/js-core-1/week-1/lesson.md
+++ b/js-core-1/week-1/lesson.md
@@ -241,7 +241,7 @@ function double(number) {
 
 To use the function we need to:
 a) _call_ it with an input and
-b) assign the returned value to a variable
+b) do something the returned value, for example assigning it to a variable
 
 ```js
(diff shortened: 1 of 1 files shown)


js-core-1/week-1/lesson.md
```diff
@@ -241,7 +241,7 @@ function double(number) {
 
 To use the function we need to:
 a) _call_ it with an input and
-b) assign the returned value to a variable
+b) do something the returned value, for example assigning it to a variable
```
ChrisOwen101:
```suggestion
b) do something with the returned value, for example assigning it to a variable
```

ChrisOwen101 (approved):

PR #465, opened 2020-07-05
Remove first item for the prepare for next class instructions as it's…

llh1:
Quick update to remove the link to the official React tutorial from week 1 homework. The tutorial only uses class components, so not relevant at this stage of the class.

### DIFF
--- a/react/week-1/homework.md
+++ b/react/week-1/homework.md
@@ -3,6 +3,5 @@
 
 ### Prepare for the next class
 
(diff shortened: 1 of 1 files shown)


alismith-dev (commented):
Looks good, but I would also suggest removing the Egghead link too. Although I like the videos, for a lot of students the fast pace is really off-putting.

If we want to keep them, we should put a limit on how many they watch, so they don't start learning about state ahead of the week 2 lesson.

PR #466, opened 2020-07-12
Fix description of what `.querySelectorAll` returns

EmilePW:
# What does this change?

Module: JS2
Week(s): 2

## Description

I've changed the description of what `.querySelectorAll` returns to be slightly more accurate - it previously said that it returns an array when it actually returns an array-like `NodeList` which doesn't implement most of the array methods. I hope now it isn't too confusing by introducing an extra concept? My worry was that students might start trying to use `.map`, etc. on a `NodeList` and be confused when that fails.

[Rendered version](https://github.com/CodeYourFuture/syllabus/blob/EmilePW-fix-querySelectorAll-desc/js-core-2/week-2/lesson.md)

## Who needs to know about this?

@ChrisOwen101

### DIFF
--- a/js-core-2/week-2/lesson.md
+++ b/js-core-2/week-2/lesson.md
@@ -85,7 +85,7 @@ document.querySelectorAll("p");
 ```
 
 Both `.querySelector` and `querySelectorAll` accept a CSS selector as an input.
-`.querySelector` selects only the first element it finds, `querySelectorAll` selects all elements (it returns an array).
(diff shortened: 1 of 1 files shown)


illicitonion:
This looks great, thanks :) It may also be worth mentioning something like:

If you want to use a `NodeList` like an Array (e.g. call `.map` on it), you can turn it into an Array like this:
```
let elementArray = Array.from(document.querySelectorAll("div"));
```

Then `elementArray` will be an `Array` containing each of the DOM elements.

EmilePW:
Thanks @illicitonion, that's a good point - I've added an `Array.from` example in too.

